PR #1, opened 2010-09-09
package.json for npm

ghost:
I added a package.json to your node-github so that people can just `npm install github` like they can for kiwi. To publish to the npm registry, just `npm publish` from the node-github project root. Thanks for the module!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,13 @@
+{
+    "name" : "github",
+    "version" : "0.0.1",
+    "descrption" : "Wrapper for the GitHub API",
+    "modules" : {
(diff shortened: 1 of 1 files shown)


meskhidzeg4-create:
owner Giorgi meskhidze

package.json
```diff
@@ -0,0 +1,13 @@
+{
```
meskhidzeg4-create, line 1: owner Giorgi Meskhidze

meskhidzeg4-create (commented):
owner Giorgi meskhidze

PR #121, opened 2014-02-03
Eliminate generated code files and their templates.

Robert-Nelson:
Most of this diff is the contents of the obsolete generated files and adding the new apidocs.js file.

The current design of generating javascript code files containing 100s of duplicate implementations of the
API handler is inefficient and unnecessary.  It also creates situations where the routes.json file can be out
of sync with the generated code.

This change drastically simplifies the code and eliminates all the duplication.

With this change generate.js is only used to generate the documentation and test code.


rngtng:
what's status of this? Would be great to get this in, as I'd like to use it with browserify too

jonschlinkert:
same, would be nice to see this updated and merged. I came here to see if anyone else had done this pr already.

PR #123, opened 2014-02-04
Fix authorizationTest

Robert-Nelson:
Removed existing authorizations and updated asserts.

Also increased the timeout to 10 seconds.


gr2m:
🤣 no idea why semantic-release thought that this was part of the latest release :) sorry for the noise

PR #166, opened 2014-07-01
Adds the upload asset release api

greggman:
Here's a first pass at adding this feature. Please tell me if you'd
like it implemented a different way. Of course I like it as is :P
but some things to consider
-   Content-Type
  
  Currently I set the content type automatically based on the name
  you pass in. This is a little magic as it assumes there's both
  a `filePath` and `name` field.
  
  It could be required to be set by the user as in
  
  ```
  github.releases.uploadAsset({
      owner: owner,
      id: releaseId,
      headers: {
         "content-type": "application/zip",
      }
      ...
  ```
  
  Similarly the size could be manually set there. But, since, based
  on the current API there's no way for us to upload the file (we
  don't get the request object back, and we can't just pass the
  data (it's arguably too big. Could be 100s of meg, more than
  fits in node's memory space), it seems like we should lookup
  the size and stream the file as it does now?
-   Timeout
  
  Wasn't sure what to do about that. Timeouts don't seem appropriate
  for uploading giant files or at least can't easily be a global
  setting.

Anyway, if you don't like this patch hopefully you can suggest a way
you'd prefer it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ github.authorization.create({
 * Search: 100%
 * Markdown: 100%
 * Rate Limit: 100%
-* Releases: 90%
+* Releases: 100%
 * Gitignore: 100%
 * Meta: 100%
 * Emojis: 100%
--- a/api/v3.0.0/releases.js
+++ b/api/v3.0.0/releases.js
@@ -326,6 +326,49 @@ var releases = module.exports = {
     };
 
     /** section: github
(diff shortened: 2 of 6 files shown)


mikedeboer:
@greggman Awesome! Apologies for the delay of my review, but all looks well and I'll merge it now. Thanks for this contribution!

ef4:
Glad to see this merged! Please publish to npm, too. Thanks.

PR #212, opened 2014-12-04
Enable filters for Organization Members

ekristen:
Allow using the filter parameter on getting organization members.


mikedeboer:
Thanks for this contribution!!

PR #216, opened 2014-12-26
Update example code library references to 'node-github'

gerrypower:


mikedeboer:
@gerrypower I don't understand... the canonical name of this package on NPM is 'github', so changing it to node-github doesn't really make sense to me. Why do you suggest this change?

PR #225, opened 2015-02-10
Update issues.create -- labels are not required

ekristen:
When creating an issue, labels are not required by GitHub API v3


mikedeboer:
@ekristen, you're very right! Thanks for this contribution!

PR #228, opened 2015-02-20
Adds code search support

zamith:
Refactors the search file to use the same function for all endpoints.


zakty:
Code root samsung gt-s5830

mikedeboer:
@zamith Thanks! There was no need to refactor things, since all the files are auto-generated.

PR #232, opened 2015-03-10
add promise to async calling

rohozhnikoff:
i work with node-github with my two pet projects
usually i use promises for async calling, because it's more useful for chaining, grouping, step-by-step code, debugging, abstract construction of queries to pseudo-pure function and other
that's why i improve my api-calls with promise-returning, based on es6-standart
want to share this

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var fs = require("fs");
 var mime = require("mime");
 var Util = require("./util");
 var Url = require("url");
+var Promise = require('es6-promise').Promise;
 
 /** section: github
  * class Client
@@ -349,7 +350,10 @@ var Client = module.exports = function(config) {
                         };
                     }
 
-                    self[section][funcName] = function(msg, callback) {
+
+
+
+                    self[section][funcName] = function(msg, callback){
                         try {
                             parseParams(msg, block.params);
                         }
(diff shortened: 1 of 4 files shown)


rohozhnikoff:
discussion in comment https://github.com/mikedeboer/node-github/issues/205

index.js
```diff
@@ -363,7 +367,20 @@ var Client = module.exports = function(config) {
                             return;
                         }
 
-                        api[section][funcName].call(api, msg, block, callback);
+                        // good description for standart: http://www.html5rocks.com/en/tutorials/es6/promises/
```
Pouja, line 370: standard*

Pouja:
Why can't an user just use Q.ninvoke? Or any other wrapper like function of Q?

rohozhnikoff:
what reason to use Q instead of existing solutions?

> Or any other wrapper like function of Q?

es6 Promise is it

jamestalmage:
@mikedeboer see #262, I think it is the better solution.

kaizensoze:
Holding off on a promise solution for now, but will probably go with #262 if any. Thanks for the PR though!

gr2m:
Hey folks, we plan to drop support for the custom Promise option altogether. If that causes any trouble for you, could you please let me know at #622? 

Thanks :)

PR #233, opened 2015-03-13
fix getNextPage when using pathPrefix

coreprocess:


mikedeboer:
Thanks for this contribution!

Issue #178, opened 2014-08-02
need wrapper for /users


dwmarshall:
`node-github` does everything I want except for a call to `/users`

mikedeboer:
`github.user.getAll()`

PR #238, opened 2015-03-29
Add getAll for users

danjm:
This should take care of Issue #178

I added a "getAll" function to user.js and added the appropriate corresponding object to routes.json.  This enables a get request to "/users", which "provides a dump of every user, in the order that they signed up for GitHub." See: https://developer.github.com/v3/users/#get-all-users

### DIFF
--- a/api/v3.0.0/routes.json
+++ b/api/v3.0.0/routes.json
@@ -2919,6 +2919,20 @@
     },
 
     "user": {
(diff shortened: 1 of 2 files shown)


raabbajam:
:+1:

mikedeboer:
Thanks for this contribution!

PR #241, opened 2015-04-19
Auto-generate documentation URLs by parsing them from a fresh grab of th...

mikedeboer:
...e GH documentation site.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -14,7 +14,9 @@
 
 var Fs = require("fs");
 var Path = require("path");
+var Url = require("url");
 
+var Async = require("async");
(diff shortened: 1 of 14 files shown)


generate.js
```diff
@@ -58,179 +69,350 @@ var main = module.exports = function(versions, tests, restore) {
         var sections = {};
         var testSections = {};
 
-        function createComment(paramsStruct, section, funcName, indent) {
```
mikedeboer, line 61: It might look like I've rewritten the entire file, but it's mostly whitespace changes (indentation).

PR #245, opened 2015-04-27
Adding support to `oauth-basic`

gullitmiranda:
More info: https://developer.github.com/v3/auth/#via-oauth-tokens


jamestalmage:
LGTM. 

Have you tested it? It would be best if you created a unit test.

PR #254, opened 2015-05-20
add spdx license string

smikes:
adding "license": "MIT" field in package.json silences
warning from newer `npm` about missing `license` field


mikedeboer:
Thanks!!

PR #255, opened 2015-05-22
Support for ETAG

sielay:
Following PR is meant to allow it inject cache managing object, that would check, what was last ETAG value of specific response. This should help avoid exceeding quotas so easily.

This implementation is based on GitHub suggestion https://developer.github.com/v3/#conditional-requests

### DIFF
--- a/index.js
+++ b/index.js
@@ -174,674 +174,696 @@ var Url = require("url");
  *          ...
  *      }
  **/
-var Client = module.exports = function(config) {
-    config.headers = config.headers || {};
-    this.config = config;
-    this.debug = Util.isTrue(config.debug);
+var Client = module.exports = function (config) {
+	config.headers = config.headers || {};
+	this.config = config;
+	this.debug = Util.isTrue(config.debug);
 
-    this.version = config.version;
-    var cls = require("./api/v" + this.version);
-    this[this.version] = new cls(this);
+	this.version = config.version;
+	var cls = require("./api/v" + this.version);
+	this[this.version] = new cls(this);
 
-    var pathPrefix = "";
-    // Check if a prefix is passed in the config and strip any leading or trailing slashes from it.
-    if (typeof config.pathPrefix == "string") {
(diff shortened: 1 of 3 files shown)


WORMSS:
Why change all spaces for tabs?

sielay:
Oh, sorry, in other projects I contribute they forced that. You mind it much?

WORMSS:
Well, it is not my project to comment, but I was looking to see what the changes were, and noticed it was EVERY LINE had changed.. so from my point of view, I could not see the proper changes.

jerone:
This will show all _real_ changes: https://github.com/mikedeboer/node-github/pull/255/files?w=1
But the style :art: changes should really be reverted.

WORMSS:
Thank you @jerone,
@sielay I like how the etag handler is external. I had tried to work with `node-github-cache` before and that had modules that required compiling which made it a no go as it had to "just work" on windows, without making sure the user had visual studio and python and gcc and a bunch of other things installed on their machine. So I tip my hat to you good sir.

sielay:
I read it as positive comment @WORMSS . I will cleanup that over the weekend and let you know.

kaizensoze:
I'd be open to merging this once all conflicts are resolved and whitespace changes undone.

kaizensoze:
I'm going to close this only because the codebase has changed so much since the PR was introduced.

gr2m:
I can’t give an estimate, but I will revise the topic of caching some time soon. You can also follow #620 if you are interested in progress on this library in general

jsg2021:
I think #673 covers this

jsg2021:
See this line: 

https://github.com/octokit/node-github/blob/master/test/integration/conditional-request-test.js#L21

gr2m:
Yeah I think that should work as a workaround for the time beeing.

I don’t want to put caching into `@octokit/rest.js` core, but I want to make it a first-class API / plugin, but can’t give an ETA on that yet

PR #258, opened 2015-06-09
Add promise support

gero3:
This adds promise support when no callback is given.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -1,0 +1,48 @@
+/*jslint browser: true, node: true, maxlen: 128 */
+/*global self */
+
+"use strict";
+
+var globalObject, hasPromiseSupport;
+
+function isFunction(x) {
+    return typeof x === "function";
+}
+
+// Seek the global object
+if (global !== undefined) {
+    globalObject = global;
+} else if (window !== undefined && window.document) {
+    globalObject = window;
+} else {
+    globalObject = self;
+}
+
+hasPromiseSupport =
+
+    globalObject.hasOwnProperty("Promise") &&
+
+    // Some of these methods are missing from
+    // Firefox/Chrome experimental implementations
+    globalObject.Promise.hasOwnProperty("resolve") &&
+    globalObject.Promise.hasOwnProperty("reject") &&
+    globalObject.Promise.hasOwnProperty("all") &&
+    globalObject.Promise.hasOwnProperty("race") &&
+
+    // Older version of the spec had a resolver object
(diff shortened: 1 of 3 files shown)


gh0stonio:
+1

promise.js
```diff
@@ -0,0 +1,48 @@
...
+hasPromiseSupport =
+
+    globalObject.hasOwnProperty("Promise") &&
+
+    // Some of these methods are missing from
+    // Firefox/Chrome experimental implementations
+    globalObject.Promise.hasOwnProperty("resolve") &&
+    globalObject.Promise.hasOwnProperty("reject") &&
+    globalObject.Promise.hasOwnProperty("all") &&
+    globalObject.Promise.hasOwnProperty("race") &&
```
jamestalmage, line 30: Why are you testing for all these methods on `Promise`? You use exactly none of them!

~~Drop lines 23-30~~.

EDIT: Drop lines 25-30

promise.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Seek the global object
+if (global !== undefined) {
+    globalObject = global;
+} else if (window !== undefined && window.document) {
+    globalObject = window;
+} else {
+    globalObject = self;
+}
```
jamestalmage, line 19: You do not need any of these tests. Just go ahead and use `global` directly. It has existed in node since the beginning. `browserify` sets the value of `global` to `window` anyways, so it will exist even in the browser.

promise.js
```diff
@@ -0,0 +1,48 @@
...
+        return isFunction(resolve);
+    }());
+
+// Export the native Promise implementation if it looks like it matches
+// the specificiation. Otherwise, return the es6-promise implementation
+// by @jaffathecake.
+if (hasPromiseSupport) {
+    module.exports = globalObject.Promise;
+} else {
+    module.exports = require("es6-promise").Promise;
```
jamestalmage, line 47: I'd prefer bluebird. [Better performance](https://gist.github.com/calvinmetcalf/b5ed2a3ab44125aa6228), and it's what all the cool kids are using.

jamestalmage:
@mikedeboer see #262, I think it is the better solution.

gero3:
See #262 for a better implementation

gr2m:
Hey folks, we plan to drop support for the custom Promise option altogether. If that causes any trouble for you, could you please let me know at #622? 

Thanks :)

PR #260, opened 2015-06-26
missing comment. msg.filePath of uploadAsset

importre:
I think that `msg.filePath` of uploadAsset is required.


importre:
oh sorry. dup #252

PR #261, opened 2015-06-30
make it work with browserify

jamestalmage:
The main issue preventing browserify usage was computed paths
in require statements: `require("./" + someVariable)`. Those have
all been eliminated.

This patch only adds browser support for `v3.0.0` (see the throwing
code in `/index.js` where it states exactly that to understand why).
Hint: it's related to computed paths again.

There were also a number of issues in `browserify-http`, and
`browserify-https` that I needed to code around to get things working:
- https://github.com/substack/https-browserify/pull/1
- https://github.com/substack/http-browserify/pull/90
- https://github.com/substack/http-browserify/issues/21
- https://github.com/substack/http-browserify/pull/10

### DIFF
--- a/api/v3.0.0/index.js
+++ b/api/v3.0.0/index.js
@@ -19,7 +19,7 @@ var error = require("./../../error");
 
 var GithubHandler = module.exports = function(client) {
     this.client = client;
-    this.routes = JSON.parse(Fs.readFileSync(__dirname + "/routes.json", "utf8"));
+    this.routes = require("./routes.json");
 };
 
 var proto = {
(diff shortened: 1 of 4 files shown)


templates/index.js.tpl
```diff
@@ -19,7 +19,7 @@ var error = require("./../../error");
 
 var GithubHandler = module.exports = function(client) {
     this.client = client;
-    this.routes = JSON.parse(Fs.readFileSync(__dirname + "/routes.json", "utf8"));
+    this.routes = require("./routes.json");
```
jamestalmage, line 22: Reading a file via the `fs` module is not well supported in browserify without a special transform.
Fortunately for json files we can just `require()` them and it works across platforms.

templates/index.js.tpl
```diff
@@ -19,7 +19,7 @@ var error = require("./../../error");
 
 var GithubHandler = module.exports = function(client) {
     this.client = client;
-    this.routes = JSON.parse(Fs.readFileSync(__dirname + "/routes.json", "utf8"));
```
jamestalmage: Doing it this way is unnecessary and does not work in the browser.

api/v3.0.0/index.js
```diff
@@ -33,8 +33,20 @@ var proto = {
...
+Util.extend(proto, require("./statuses"));
+Util.extend(proto, require("./pullRequests"));
+Util.extend(proto, require("./repos"));
+Util.extend(proto, require("./user"));
+Util.extend(proto, require("./events"));
+Util.extend(proto, require("./releases"));
+Util.extend(proto, require("./search"));
+Util.extend(proto, require("./markdown"));
+Util.extend(proto, require("./gitignore"));
+Util.extend(proto, require("./misc"));
```
jamestalmage, line 50: This is still generated code - see the updates to `generate.js`

jamestalmage:
Anyone wishing to use my patched version can install it with the following command:

`npm install --save jamestalmage/node-github#browserify-compatibility`

You can then use it like you normally would server side. (you still just `require("github")`).

olivierbeaulieu:
@mikedeboer Any chance we can get this work re-started? I would _love_ Browserify support for this package. Any reason why this was never merged? I don't mind doing the work to make it mergeable again.

kaizensoze:
Yeah I've been backlogged with stuff, but it's on my radar.

PR #262, opened 2015-06-30
YAPPR - Yet Another Promise PR

jamestalmage:
this simplifies the initial implementation of Promise support by @gero3, and allows for pluggable promise implementations. If a global `Promise` function exists, it will use that. Optionally, users can supply a promise implementation via `config.Promise`:

``` javascript
var GitHubApi = require('github');

var gh = new GitHubApi ({
  version: '3.0.0',
  Promise: require('bluebird')
});
```

This keeps the dependency list small, and gives people the freedom to choose their favorite Promise library (there have been at least 3 different libraries recommended in the issues so far). Most promise libraries are pretty huge, and now that I [have this working in the browser](https://github.com/mikedeboer/node-github/pull/261), I would be concerned about forcing large dependencies on people. This also has potential benefits for browser frameworks with existing promise implementations (I'm thinking specifically of [angular.js $q](https://docs.angularjs.org/api/ng/service/$q)).

see #258, #232

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,6 +179,7 @@ var Client = module.exports = function(config) {
     config.headers = config.headers || {};
     this.config = config;
     this.debug = Util.isTrue(config.debug);
+    this.Promise = config.Promise || config.promise || Promise;
 
     this.version = config.version;
     var cls = require("./api/v" + this.version);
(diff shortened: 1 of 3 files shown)


hofan41:
Just stumbled in, it appears that the lead maintainer for this project has gone on hiatus, and I am itching for this feature. 

When can we expect @mikedeboer to be back and responding? If not, should a new fork be created?

2color:
@jamestalmage There's effort to merge some of the PRs in this fork:
https://github.com/kaizensoze/node-github

I'd suggest to make the PR there.

jamestalmage:
I'm not going to have time for that in the near future.

codydaig:
:+1: for this! I currently just wrap the calls in helper functions that return a new promise and this would simplify the code a lot!

kaizensoze:
@codydaig Couldn't you do something like

```
var Promise = require("bluebird");
var GitHubApi = Promise.promisifyAll(require("github"));
```

kaizensoze:
@codydaig Fair enough

thejohnfreeman:
Thanks for merging!

gr2m:
Hey folks, it's been a year and things have changed. We plan to drop support for the custom Promise option altogether. If that causes any trouble for you, could you please let me know at #622? 

Thanks :)

PR #272, opened 2015-08-05
pull-request / update : title not required

valmiRe:
when I update a PR, the title is optional for example when I only want to update the state


valmiRe:
fixed here https://github.com/kaizensoze/node-github/commit/29a00f0160f739c39e2660881357dfe6b691fd00
thanks @kaizensoze

kaizensoze:
:+1:

PR #307, opened 2016-01-11
adds semicolon for consistency

ClimbsRocks:
just adding a single semicolon to be consistent with the rest of the doc


kaizensoze:
Doc has changed recently

PR #311, opened 2016-01-29
Fix the generation of documentation links after github changed the st…

ankon:
…ructure


kaizensoze:
Doc generation has changed

PR #318, opened 2016-03-24
Pass res through to callback when reqeusts have a non 200 response

burtonjc:
We log out requests that we make to GitHub. Currently if the request has a non 200 response, we do not have a handle to the response object. Ironically, it is in that case that we most care about having it.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -691,7 +691,7 @@ var Client = module.exports = function(config) {
                 });
                 res.on("end", function() {
                     if (res.statusCode >= 400 && res.statusCode < 600 || res.statusCode < 10) {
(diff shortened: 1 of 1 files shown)


kaizensoze:
Headers are included in the failure case if that helps: https://github.com/kaizensoze/node-github/pull/93/files (merged into this repo)

burtonjc:
That isn't enough for us. We actually need the res body to see the error message. I updated my PR to get rid of the merge conflict.

burtonjc:
I'm actually going to close this PR an reopen it from a different head branch. I'll link back to it.

burtonjc:
New PR open here: https://github.com/mikedeboer/node-github/pull/347

PR #325, opened 2016-04-28
Feature/add repos update team

carnun:


carnun:
Just thought I'd add this missing API call since I needed it for one of my projects. Hope I've added it the endpoint to the right place.

PR #346, opened 2016-06-03
feat(multiple-assignees): Adding support for updating multiple assignees

burtonjc:
As per: https://developer.github.com/v3/issues/#create-an-issue
and: https://developer.github.com/v3/issues/#edit-an-issue

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -1579,6 +1579,13 @@
                     "invalidmsg": "",
                     "description": "Login for the user that this issue should be assigned to."
                 },
+                "assignees": {
+                    "type": "Json",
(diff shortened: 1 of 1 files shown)


kaizensoze:
The `assignees` param is preview only at the moment.

burtonjc:
True, you just can't use `node-github` to update `assignees` without this change. Up to you though. No hard feelings if you don't want this in yet.

kaizensoze:
I'm going to leave this out for now only because `node-github` doesn't handle preview stuff at the moment.

burtonjc:
Cool, thanks for all the quick reviews!

PR #347, opened 2016-06-03
feat(error-callback): Pass res through to callback when reqeusts have a non-200 response

burtonjc:
We need access to the response body so we can handle errors in the most appropriate way possible.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -691,7 +691,7 @@ var Client = module.exports = function(config) {
                 });
                 res.on("end", function() {
                     if (res.statusCode >= 400 && res.statusCode < 600 || res.statusCode < 10) {
-                        callCallback(new error.HttpError(data, res.statusCode, res.headers));
+                        callCallback(new error.HttpError(data, res.statusCode, res.headers), res);
                     } else {
                         res.data = data;
                         callCallback(null, res);


burtonjc:
Reopened from https://github.com/mikedeboer/node-github/pull/318

kaizensoze:
Can you give an example of a particular error and what you need from `res` that you can't get from the existing error object?

burtonjc:
Validation errors in GitHub, for example, will send a response body like:

```
{
  "message": "Validation Failed",
  "errors": [
    {
      "value": "doesn'texist",
      "resource": "Issue",
      "field": "assignee",
      "code": "invalid"
    }
  ],
  "documentation_url": "https://developer.github.com/v3/issues/#edit-an-issue"
}
```

Without getting that response body, all you have to work with is the `Status: 422 Unprocessable Entity` response header. That doesn't actually give you information on what happened though. If we get an error from GitHub, we need to know why it was that the request failed so we can handle it appropriately.

kaizensoze:
That's exactly what the error message is though:

``` javascript
github.issues.edit({
    user: "kaizensoze",
    repo: "test2",
    number: 2,
    title: "sunkist2",
    assignee: "ladjlfjasdfaasfka" // invalid asignee
}, function(err, res) {
    console.log(err.message);
});
=> 
{"message":"Validation Failed","errors":[{"value":"ladjlfjasdfaasfka","resource":"Issue","field":"assignee","code":"invalid"}],"documentation_url":"https://developer.github.com/v3/issues/#edit-an-issue"}
```

burtonjc:
Oi, ha. Not sure how I missed that. Sorry for the bother. I'll close this one out.

PR #348, opened 2016-06-03
fix(redirects): Configuration to disable redirects

### LINKED ISSUE #345: Bug: Redirects are broken for all non-GET requests
The `follow-redirects` module takes any redirect and replays it as a `GET` to the redirected url. GitHub also redirects `POST`s, `PUT`s, `PATCHE`s, and `DELETE`s. This means that if you update an issue and the request gets redirected, the redirect is made as a `GET` and you simply get the non-updated issue back.

I would support either removing it, or making it configurable in `node-github` so there is a way to turn it off.

burtonjc:
closes https://github.com/mikedeboer/node-github/issues/345

Our app is currently broken on the latest version of `node-github` because when we get redirects from GitHub for non-`GET` requests, `node-github` replays them as `GET`s. We were already handling all the redirects on our side, so being able to disable it in `node-github` is enough for us.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -676,7 +676,8 @@ var Client = module.exports = function(config) {
             console.log("REQUEST: ", options);
 
         function httpSendRequest() {
-            var req = require('follow-redirects/' + protocol).request(options, function(res) {
+            var reqModule = self.config.followRedirects === false ? protocol : 'follow-redirects/' + protocol;
+            var req = require(reqModule).request(options, function(res) {
                 if (self.debug) {
                     console.log("STATUS: " + res.statusCode);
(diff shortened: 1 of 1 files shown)


kaizensoze:
I'll try to have this and your other PRs tested and hopefully merged by this weekend.

burtonjc:
Awesome. Thank you!

burtonjc:
Thanks for the quick responses!

gr2m:
Hey @burtonjc, I’m working on removing the `followRedirects` option. Redirects will always be followed, but the method forward should work correctly. If you could test your app with https://github.com/octokit/node-github/pull/659, that’d be much appreciated.

If you see any problem with removing the option, please let me know, this will not be released for some time

PR #357, opened 2016-06-10
allow authorization header in request

valmiRe:
I have a multi user application and I want to avoid calling the authenticate() method on every request
This will work only with the token auth type, for a global solution, we could pass an _authenticate_ object as a parameter to the requests, this will require a very little modification in the code

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -34,6 +34,7 @@
             "Status"
         ],
         "request-headers": [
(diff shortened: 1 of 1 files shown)


kaizensoze:
Can you just capitalize it to match the other headers?

valmiRe:
ok I've made the modification to match other headers

PR #358, opened 2016-06-12
Fixed Browserify compatibility

olivierbeaulieu:
Browserify is unable to follow dynamically generated paths. By swapping out the `fs.readFileSync` in favour or a `require` call, this makes the library work nicely with Browserify.

Some changes were needed in the http requests code. Are http and https the only supported protocols? I could not find any documentation on this.

I am unable to run the tests, even on master. Can you provide some more explanation on how I should run them? `npm test` throws errors, even on master.

Aside from that, everything seems to work fine.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -74,7 +74,7 @@ var Client = module.exports = function(config) {
     this.config = config;
     this.debug = Util.isTrue(config.debug);
 
-    this.routes = JSON.parse(fs.readFileSync(__dirname + "/routes.json", "utf8"));
+    this.routes = require("./routes.json");
 
     var pathPrefix = "";
     // Check if a prefix is passed in the config and strip any leading or trailing slashes from it.
@@ -413,7 +413,7 @@ var Client = module.exports = function(config) {
                 callback(null, ret);
         });
     }
-    
+
     /**
      *  Client#getAllPages(method, options, callback) -> null
      *      - method (Function): github client method to be called recursively
@@ -678,8 +678,24 @@ var Client = module.exports = function(config) {
             console.log("REQUEST: ", options);
 
         function httpSendRequest() {
-            var reqModule = self.config.followRedirects === false ? protocol : 'follow-redirects/' + protocol;
-            var req = require(reqModule).request(options, function(res) {
+            var requestModule;
+
+        	if(self.config.followRedirects === false) {
+        		if(protocol === 'https') {
+        			requestModule = require('https');
+        		} else {
+        			requestModule = require('http');
+        		}
+
+        	} else {
+        		if(protocol === 'https') {
+        			requestModule = require('follow-redirects/https');
+        		} else {
+        			requestModule = require('follow-redirects/http');
+        		}
+        	}
+
+            var req = requestModule.request(options, function(res) {
                 if (self.debug) {
                     console.log("STATUS: " + res.statusCode);
                     console.log("HEADERS: " + JSON.stringify(res.headers));
--- a/lib/index.js
+++ b/lib/index.js
@@ -74,7 +74,7 @@ var Client = module.exports = function(config) {
     this.config = config;
     this.debug = Util.isTrue(config.debug);
 
-    this.routes = require("./routes.json");
+    this.routes = Object.assign({}, require("./routes.json"));
 
     var pathPrefix = "";
     // Check if a prefix is passed in the config and strip any leading or trailing slashes from it.


lib/index.js
```diff
@@ -678,8 +678,24 @@ var Client = module.exports = function(config) {
             console.log("REQUEST: ", options);
 
         function httpSendRequest() {
-            var reqModule = self.config.followRedirects === false ? protocol : 'follow-redirects/' + protocol;
```
kaizensoze, line 681: Does this break as-is with browserify?

olivierbeaulieu, line 681: Yes - Browserify/webpack both analyze the require paths statically at build time, so they both can't follow paths constructed at runtime.

kaizensoze:
Do you have a simple browserify and webpack example to test against? I was testing this with both and getting errors.

olivierbeaulieu:
Right, I just figured out that an additional hack needs to be baked in. One of the lib's dependencies is querying `process.version`, which is undefined in Browserify. I hard-coded it to `5.10.1` in my project, which made this work.

Let me figure out an elegant way to shim this out and I'll add it to my PR.

kaizensoze:
Ok cool, sounds good.

canyousayyes:
+1 to this PR, but with different use case. 
I am using this package to run codes in AWS Lambda. (AWS Lambda is a cloud service to execute nodejs scripts).
And I am using serverless-babel to manage my application. ( https://github.com/serverless/serverless-runtime-babel )

When deploying our application, babel will resolve the `__dirname` locally, meaning the compiled script will become something like `/Users/myname/app/node_modules/github/lib/routes.json`.
This is ok is local environment, but it will cause error when running in the cloud service (since there's no such file in that path).

When using `require()`, the problem will be solved.

BuggieLiam (approved):

stephenmathieson:
What's the status of this PR? I'm also trying to use this package in a lambda function (which uses browserify to reduce the lambda's .zip size), but am unable to.

Any way we can get this conflict resolved and merge this?

zeke:
> What's the status of this PR?

Folks at GitHub are working to improve this module, and we recognize browser compatibility as an important aspect. See https://github.com/octokit/discussions/issues/1

Our first priority is getting a working test suite, but after that we can definitely focus on supporting browserify.

gr2m:
@olivierbeaulieu did  you ever manage to use the browserified version of GitHub in browser?

When I do the following:

```shell
browserify lib/index.js --standalone GitHub -o github.js
cat github.js | pbcopy
```

Then open any website like https://example.com and paste the code of `github.js` into the console, the `GitHub` object becomes available. But when I run this

```js
var github = new GitHub()
github.repos.get({user: 'octocat', repo: 'Hello-World'})
```

Then nothing happens, and the method does not return a promise, it returns undefined. When I run this code

```js
github.repos.get({user: 'octocat', repo: 'Hello-World'}, console.log)
```

It logs

```json
{
  "code": "400",
  "status": "Bad Request",
  "message": "Empty value for parameter 'user': undefined"
}
```

stephenmathieson:
@gr2m my fork (`@sm-forks/github`) has this patch applied and works just fine when browserified (for the server, with the `--node` option). I'm using it in an app currently and all seems to be fine

IIRC a bit more work needs to happen for this to work _in the browser_ tho, and imagine it won't have perfect parity with the node version

stephenmathieson:
@kaizensoze @zeke it looks like you're now actively working on this project. Any chance we could get this merged soon?

gr2m:
@stephenmathieson I don’t think the PR is good to merge as it doesn’t work in the browser. We are working on this but will probably do drastic changes and it will take a while

gr2m:
<details><summary>comment addressed</summary>
@kaizensoze to be fair, plan is to move the repo as far as I know, so all issues & PRs will be moved, too :) 
</details>

kaizensoze:
Yeah my comment was technically inaccurate. It'll be moved to octokit. Comment edited.

stephenmathieson:
Please reconsider merging this. 

This patch doesn't claim to fix/add browser support. Instead, it enables users to browserify their apps before shipping off to Lambda (or whatever other FaaS provider). It addresses the problem for my use case (and many others, hence this PR and the several issues). It seems like a trivial change which doesn't block anyone from making browser-compatible updates.

However, as previously stated, my fork works for now, but I don't think forcing people to re-publish to npm/rely on random forks in production is great UX.

gr2m:
https://github.com/octokit/node-github/pull/637 should fix the problem, if you could give it a try, that'd be much appreciated

I've also started looking into browserify/webpack compatibility in general. You can follow updates at
https://github.com/octokit/node-github/issues/635 and general updates at https://github.com/octokit/node-github/issues/620

PR #369, opened 2016-06-21
allow empty path for `getContent`

rip-tyang:
Empty path in getContent should be allowed.
It is supposed to return the contents of the root path.

Now it will throw an error according to [the code here](https://github.com/mikedeboer/node-github/blob/master/lib/index.js#L161).


kaizensoze:
Thanks.

PR #404, opened 2016-08-02
Allow for passing some required parameters as null

fheisler:
The new "update branch protection" endpoint requires passing both `required_status_checks` and `restrictions` objects, although both can be passed as nulls.

This PR adds an `allow-null` that can be set for specific parameters, similar to `allow-empty` for required but empty strings, and sets `allow-null` to true for `required_status_checks` and `restrictions`.

I have not added/updated any tests, as it looks like they're automatically generated. I ran `generate.js`, but it didn't pick up any changes to include.


kaizensoze:
Thanks!

PR #413, opened 2016-09-03
Updated getForUser for Organizations

leroyjware:
- Update getForUser to use /user/orgs rather than /users/:user/orgs

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -2678,7 +2678,7 @@
         },
 
         "get-for-user": {
-            "url": "/users/:user/orgs",
+            "url": "/user/orgs",
             "method": "GET",
             "params": {
                 "$user": null,


leroyjware:
I had to update this in my node_modules folder to list my organizations. Is this a bug?

kaizensoze:
There's `users.get-orgs`, which corresponds to `/user/orgs`, and then there's `orgs.get-for-user` which corresponds to `/users/:user/orgs`. Technically `users.get-orgs` should be under orgs if you were strictly mapping to the github api docs (https://developer.github.com/v3/orgs/), which this api does for the most part, with `users.get-orgs` being an exception, because the `*-for-user` endpoints map to `/users/:user/*`, but in this case it's `/user/orgs` so if you moved it it would be `orgs.get-orgs` which doesn't make as much sense semantically. _However_, I do see that the descriptions for both `users.get-orgs` and `orgs.get-for-user` need to be changed to avoid additional confusion. Namely:

```
users.get-orgs
Get your organizations -> List organizations for the authenticated user.
```

```
orgs.get-for-user
List user's organizations -> List public organization memberships for the specified user.
```

Does that make sense?

leroyjware:
Just tested users.getOrgs. That worked. I appreciate it. I'll close this.

kaizensoze:
Cool, I'll reference this issue when fixing the descriptions. Is the orgs.getForUser endpoint not working for you though?

leroyjware:
Yea, looks like something changed on Github. When I look at the documentation, Github is using /user/orgs: https://developer.github.com/v3/orgs/#list-your-organizations

I could have also probably solved by using /v3. I didn't try that either.

kaizensoze:
Right, but there's also https://developer.github.com/v3/orgs/#list-user-organizations further down. You just need to provide a username since it's not for the authenticated user.

kaizensoze:
I should really add a getUserOrgs example though using both endpoints.

kaizensoze:
Added example: https://github.com/mikedeboer/node-github/blob/master/examples/getUserOrgs.js

You don't have to specify the version since it defaults to the latest (v3).

leroyjware:
Nice. Thanks! :)

PR #419, opened 2016-09-16
Add support for integration/installations

cyberdelia:
Add support for integration/installations, as described here: https://developer.github.com/early-access/integrations/authentication/


kaizensoze:
This is merged, but I'm going to wait a bit before pushing to npm.

PR #426, opened 2016-10-04
add integrations/get-installations to routes.json

ramonsnir:
See https://developer.github.com/v3/integrations/#find-installations


kaizensoze:
Thanks!

PR #433, opened 2016-10-09
change user to owner in endpoints

### LINKED ISSUE #430: consider replacing user with owner where appropriate
Not being able to directly copy/paste from the github api docs without having to make edits is annoying.

kaizensoze:
Change user to owner in endpoints where appropriate to match official
github api urls and make it easier to copy/paste.

resolves #430

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 5.0.0
+
+Breaking changes:
+  * Replace user with owner in endpoints
+
 ## 4.1.0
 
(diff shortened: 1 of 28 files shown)


denis-sokolov:
This is a positive change. It is well documented in the changelog and you have released a new major version, great. The last remaining step you could have chosen to take is to not break backwards compatibility and accept the old parameter, maybe with a deprecation notice. Then we would not need to do this annoying change in our code. :)

kaizensoze:
Ah good point. Sorry about that.

PR #439, opened 2016-10-12
use lowercase `accept` header

christophwitzko:
Hi everyone,

using the new `integration` authentication will cause setting the accept header twice (`Accept` and `accept`). This is a quick fix for that issue. We should still think about normalizing all headers.

Best,
Christoph


finnp (approved):

kaizensoze:
Thanks!

christophwitzko:
wow, that was fast! thanks @kaizensoze ☺️

PR #442, opened 2016-10-12
Use body by default for non head/get/delete requests

finnp:
Hey,

I noticed that for `github.repos.createStatus` the parameter were send as query parameter instead of the post body like it used to be. 

This is because of this https://github.com/mikedeboer/node-github/commit/acc62b29007f97736c9582ac67a204fe97beee79 recent change, that required the `msg` to have a `body` attribute, which is not the case here. I readded `("head|get|delete".indexOf(method) === -1)`, so it will use the body for `POST` requests by default.

Best,
Finn

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -601,7 +601,7 @@ var Client = module.exports = function(config) {
         var self = this;
         var method = block.method.toLowerCase();
         var hasFileBody = block.hasFileBody;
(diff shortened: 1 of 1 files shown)


kaizensoze:
Thanks. I'll try to get this fix out asap.

finnp:
Thank you!

PR #446, opened 2016-10-19
reject promise for missing parameters

finnp:
Hey,

right now the library doesn't throw an error if I specify the wrong parameters and using Promises. This PR adds the rejection that should happen in that case.

Best,
Finn


kaizensoze:
Sweet, thanks!

PR #448, opened 2016-11-06
fix typo in validation regex for get-org-membership

ZauberNerd:
Looks like this should be an "enum" with only two valid values?
https://developer.github.com/v3/orgs/members/#list-your-organization-memberships


kaizensoze:
Yep, thanks!

kaizensoze:
Pushed as 6.0.2.

PR #451, opened 2016-11-07
explicitly list enum values in routes

ZauberNerd:
I will rebase #449 after this has been merged and use the enums instead
of parsing the regular expressions.

Maybe it would also be a good idea to change the parameter validation
logic to use the enums to generate regular expressions and remove the
"validation" property from the json (in those cases where there are enum
values) completely to not have the same information represented twice
in this file.

There's only two cases where an enumeration is not enough and a regular
expression is still needed:
- `milestone` in "/repos/:owner/:repo/issues"
	`^([0-9]+|none|\\*)$`
- `position` in "/projects/columns/cards/:id/moves" and
	"/projects/columns/:id/moves"
	`^(first|last|after:\\d)$`

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -200,6 +200,7 @@
                 "validation": "^(asc|desc)$",
                 "invalidmsg": "asc or desc, default: desc.",
                 "description": "",
+                "enum": ["asc", "desc"],
                 "default": "desc"
             },
             "since": {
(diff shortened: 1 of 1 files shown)


kaizensoze:
I'm hesitant in removing the validation param because of those edge cases. I'll think about it.

PR #459, opened 2016-11-08
gitdata.createReference update the description for doc

Lellansin:
https://developer.github.com/v3/git/refs/#create-a-reference


kaizensoze:
Sorry for some reason I thought this was an issue instead of a PR. Globally changing the ref param doesn't work due to it being so inconsistent across endpoints. Ex: https://developer.github.com/v3/repos/contents/#get-archive-link. I've updated the ref params for get-archive-link and create-reference. 9be0e4a5b6e1a656ff86c4aba84ca95da90f3258, 2d55b53c16ac2e4416b5504b2543715952058f56

PR #464, opened 2016-11-19
Storing GitHub request ID and retry-after header

jeffwilcox:
This adds two additional headers to those available in
the response metadata:

X-GitHub-RequestId: ID that is helpful to GitHub support if needed

Retry-After: to address abuse rate limits as documented in the GitHub
documentation at https://developer.github.com/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits


kaizensoze:
Thanks!

PR #474, opened 2016-12-07
Provides correct values for moveProjectsCard "position" parameter

ahwitz:
As per https://developer.github.com/v3/projects/cards/#move-a-project-card

Valid values with the Node module for the "position" parameter are currently `first|last|after:`
Putting those into `github.projects.moveProjectsCard` passes internal validation, but the API returns:
```
Move failed: {"message":"Invalid request.\n\nfirst does not match /^(?:top|bottom|after:\\d+)$/.", "documentation_url":"https://developer.github.com/v3"}
```

I've changed the validation property in `routes.json` as well as the description to match the API, and the `top|bottom` values work fine. 

I have *not* run `node lib/generate.js` as it changed a ton in doc/apidoc.js with the 7.0.1 switch, and I'm not sure how to verify that all that should be correct. `npm test` also only passes 11 of 300+, but that is also the case for the current HEAD.

Let me know if anything else is necessary for this.

### DIFF
--- a/doc/apidoc.js
+++ b/doc/apidoc.js
@@ -31334,7 +31334,7 @@ github.projects.getRepoProjects({ ... });
  * @apiGroup projects
  *
  * @apiParam {String} id  
- * @apiParam {String} position  Can be one of first, last, or after:<column-id>, where <column-id> is the id value of a column in the same project.
+ * @apiParam {String} position  Can be one of top, bottom, or after:<column-id>, where <column-id> is the id value of a column in the same project.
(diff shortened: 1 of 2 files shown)


kaizensoze:
Thanks. Yeah tests have been broken for a while. Would you be able to run node lib/generate.js, revert the doc changes, and commit all other changes?

ahwitz:
Added another commit - if that's what you're looking for, I can close this, combine the commits, and start a new PR.

kaizensoze:
Perfect, thanks!

PR #493, opened 2017-01-14
Only set generated Array body properties to string[]

### LINKED ISSUE #492: Wrong type for comment body in Typescript defs
Function definition for [`createComment`](https://github.com/mikedeboer/node-github/blob/263e57f411f00fe90978da16761efbd67b8dd3ba/lib/index.d.ts#L2660):

```
createComment(params: Github.IssuesCreateCommentParams, callback?: Github.Callback): Promise<any>;
```

With [`Github.IssuesCreateCommentParams`]() defined as:

```
export type IssuesCreateCommentParams =
    & Owner
    & Repo
    & Number
    & {
      body: string[];
    };
```

but body should be a string, not an array according to [Github API docs](https://developer.github.com/v3/issues/comments/#create-a-comment).

Not sure how to update this?

jimmidyson:
Fixes #492


lib/routes.json
```diff
@@ -428,6 +428,13 @@
                 "validation": "",
                 "invalidmsg": "",
                 "description": "OAuth token"
+            },
+            "body": {
```
kaizensoze: This is a duplicate of the one defined on line 260.

lib/routes.json
```diff
@@ -5347,7 +5360,13 @@
                 "$owner": null,
                 "$repo": null,
                 "$sha": null,
-                "$body": null,
+                "body": {
```
kaizensoze: This can be $body since it matches the global definition of type: String, required: true.

lib/generateTypes.js
```diff
@@ -44,6 +43,11 @@ function replaceArrayTypes(type, name) {
         case "reviewers":
         case "comments":
             return "string[]";
+        case "body":
+          if (type === "Array") {
```
kaizensoze: Should this just be `return "string";`?

jimmidyson: There are a number of routes that use body for the property name which is sent as the request body, rather than body as a parameter name. I'll look to see if I can tidy that up later as I think some of those types are wrong too.

kaizensoze: Ok thanks!

jimmidyson:
I've renamed other properties named `body` to a more relevant name, as well as tidying up a few other related bits in routes. I can't get tests to run but looking at the generated tests it looks like they don't run anyway? e.g. I see `"Number"` values for `Number` type properties which throw validation errors?

jimmidyson:
Any chance of a review & hopefully merge?

kaizensoze:
Yep sorry for the delay!

jimmidyson:
Thank you!

PR #501, opened 2017-02-03
fix typescript union types in template and generation scripts

### LINKED ISSUE #524: Typescript *.d.ts files are generated incorrectly
Basically, the generate script generates invalid typescript declaration files which prevent compilation at worst (and produce dozens of "warning errors" that clog up the terminal at best).

The issue is described in more detail in the pull request I created: #501

leviathanbadger:
Reason for change: the Typescript language does not support leading or trailing operators in type declarations. The | and & operators are binary, and must be used as an infix between two other types.

Currently, the .d.ts.tpl and .js.flow.tpl mustache templates ignore this requirement and produce erroneous typings, which make it impossible to use typescript in this repository. Example:

```
   export type Auth =
      | AuthBasic
      | AuthOAuthToken
      | AuthOAuthSecret
      | AuthUserToken
      | AuthJWT;
```

...should be:

```
   export type Auth =
        AuthBasic
      | AuthOAuthToken
      | AuthOAuthSecret
      | AuthUserToken
      | AuthJWT;
```

Note that mustache is a very low-level templating engine, and doesn't natively support repeating a value every time but the first/last. So, to correct this error, we must pass the values to the templating engine as objects rather than directly as strings.

So, instead of requestHeaders being a `string[]`, we must change it to `{ header: string, first: boolean }[]`.

I don't know what the js.flow file is, but it looks a lot like typescript, so I figured it wouldn't hurt to apply the same change there.

Closes #524

### DIFF
--- a/lib/generateTypes.js
+++ b/lib/generateTypes.js
@@ -118,6 +118,9 @@ var main = module.exports = function(languageName, templateFile, outputFile) {
 
     var defines = routes.defines;
     var requestHeaders = defines["request-headers"];
+    requestHeaders = requestHeaders.map(function(header, index) {
+        return { header: JSON.stringify(header), first: (index == 0) };
+    });
     delete routes.defines;
 
     Util.log("Generating " + languageName + " types...");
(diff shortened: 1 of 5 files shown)


ericanderson:
This shouldn't be needed with modern versions of TS. It works fine.

leviathanbadger:
The question is, is node-github intended to support older versions of the language? I'll update the PR if you feel that it's worth it, otherwise I'll just close it.

It may be worth mentioning that [the latest publicized specification](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.8.6) for TypeScript does not support that syntax. It is, admittedly, for Typescript 1.8, so... someone dropped the ball in the interim.

ericanderson:
Typescript 2 has been out for almost a year now. It's safe to assume this change isn't needed

PR #506, opened 2017-02-18
Updated readme examples to es6

amilajack:


kaizensoze:
I'd accept this if it was just es6 changes, but indentation has been changed from 4 to 2 spaces, `--save` has unnecessarily been added to the install command, and the basic auth example was removed. Are double quotes not allowed in es6?

amilajack:
These are just conventions that I've seen. They're not required

jeffwilcox:
@amilajack any chance you'd be willing to update your pull request to address the feedback (`--save`, etc.)?

As you're a first-time contributor we'd love to help you get the commit in.

While I also prefer single quotes these days, I think making the bare minimum changes - `const`, `=>`, etc. is ideal for now instead of making the broader changes.

Let us know what you'd like to do!

amilajack:
@jeffwilcox So I should remove the `--save` right?

jeffwilcox:
@amilajack yes, remove the `--save` for now, and also the spacing/quote changes as requested.

zeke:
My take:

`--save` should always be included in installation snippets for non-global modules. This prevents "works on my machine" issues where one person has the right modules, but others don't.

ES6+ syntax is nice, but the good thing about the existing README is that the code will work as-is in a browser without transpilation using something like babel. Granted this module doesn't currently support browsers, but it will eventually: #358 

Regarding syntax, I think we should avoid bike-shedding by using `standard` and `standard-markdown`.

kaizensoze:
I'm going to close this to avoid bikeshedding.

PR #514, opened 2017-03-02
Support for operations with empty parameters in Typescript and Flow

### LINKED ISSUE #513: users.get() fails when executed
The following code is supposed to return the current user object:

```javascript
github.users.get()
            .then((userData) => {
                // do something
            })
            .catch((err) => {
                // process the error
            });
```

But it always invokes the catch with the following error:
`'TypeError: Cannot read property 'body' of undefined' at node_modules/github/lib/index.js:621:50`

Currently using version `9.0.0`, but the same happened with the previous version

PabloZaiden:
Fixes #513


PabloZaiden:
@mikedeboer any news on this?

PR #516, opened 2017-03-06
Org blocks

williamkapke:
The Org Blocking API is the same as the User Blocking API, so I basically just copied commit 8413595a3c5071addef14d679cc51a8c35e54f44 and added the `org` parameter.

Also: I just found (not surprising) that API supports pagination when listing the block users. So, I added `page` and `per_page` params.

https://developer.github.com/v3/orgs/blocking/

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1144,6 +1144,23 @@ declare namespace Github {
     & Org
     & Id
     ;
+  export type OrgsGetBlockedUsersParams =
+    & Org
+    & Page
+    & PerPage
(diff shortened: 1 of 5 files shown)


kaizensoze:
Awesome, thanks!

williamkapke:
That's great news! [It's been working great](https://github.com/williamkapke/orgbot) so far! LMK when you publish so I can remove the temp reference to my branch.

kaizensoze:
Should be published as 9.2.0 now.

PR #517, opened 2017-03-06
set accept header on pagination

williamkapke:
When using the [Preview APIs](https://github.com/mikedeboer/node-github#preview-apis) (which nicely sets the custom `Accept` header) and then trying to paginate using `getNextPage`, `getPreviousPage`, and `getLastPage` fails because the custom `Accept` header does not get set on these calls.

Fortunately, GitHub provides a `X-GitHub-Media-Type` **response** header that echos the requested type. This PR uses this to re-create the `Accept` header on the paged requests.

Example `X-GitHub-Media-Type: github.giant-sentry-fist-preview; format=json`

**Non-preview API calls** return `X-GitHub-Media-Type: github.v3; format=json`... which works just fine also.

related: #229

If this PR is not desirable, I'll note for others that you can get this response header returned in your  meta objects by adding the `X-GitHub-Media-Type` value to the `responseHeaders` array like this:
```js
const GitHubApi = require('github')
const github = new GitHubApi({})

github.responseHeaders.push('x-github-media-type')
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -478,6 +478,7 @@ var Client = module.exports = function(config) {
             callback = headers;
             headers = null;
         }
+        headers = applyAcceptHeader(link, headers);
         return getPage.call(this, link, "next", headers, callback);
     };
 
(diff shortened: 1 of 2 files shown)


jeffwilcox:
Mind blown. Wonder if this is specific to the user blocking API you're using... I've been paginating a bunch on preview collaborator APIs and not had an issue using the new `affiliation` field in preview... hmm.

PR #521, opened 2017-03-10
Collaborators preview API "direct" affiliation

### LINKED ISSUE #520: Organization collaborators should also support 'direct' enum value (korra-preview)
Hi,
The organization collaborator work submitted in late 2016 does not allow the `affiliation` value of `direct`, although the API does.

Location of the route/API schema:
https://github.com/mikedeboer/node-github/blob/master/lib/routes.json#L5180

GitHub preview API doc: https://developer.github.com/v3/repos/collaborators/#list-collaborators

Should be something similar to:
```
 "affiliation": {
                    "type": "String",
                    "required": false,
                    "validation": "^(outside|all|direct)$",
                    "invalidmsg": "outside, all, direct, default: all",
                    "description": "Filter collaborators returned by their affiliation.",
                    "enum": [
                        "outside",
                        "all",
                        "direct"
                    ],
                    "default": "all"
```

Changes were just to add 'direct'.

Happy to make the change but also hoping perhaps it would be quick for someone else to make!

jeffwilcox:
The Collaborators preview API also supports an affiliation value
of "direct" in addition to "all" and "outside". This adds the
value to the route and regenerates the API docs.

Fixes #520

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1786,7 +1786,7 @@ declare namespace Github {
     & Page
     & PerPage
     & {
-      affiliation?: "outside"|"all";
+      affiliation?: "outside"|"all"|"direct";
     };
   export type ReposCheckCollaboratorParams =
     & Owner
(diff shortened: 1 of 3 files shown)


kaizensoze:
Thanks!

PR #547, opened 2017-06-08
Add family option

tarmolov:
It's a vital thing for ipv6 only servers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ $ npm install
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


zeke:
Thanks, @tarmolov. Can you explain what this option does and when it should be used?

tarmolov:
https://nodejs.org/api/net.html
> family <number\>: Version of IP stack, can be either 4 or 6. Defaults to 4.

PR #557, opened 2017-07-19
add NO_PROXY environment variable handling

goodefroi:
Title says it all. Currently only HTTP_PROXY and HTTPS_PROXY are taken into account but NO_PROXY is ignored. This might be important for users behind a corporate firewall connecting to a corporate GitHub that is located inside the corporate network.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -659,11 +659,27 @@ var Client = module.exports = function(config) {
         var agent = undefined;
         if (this.config.proxy !== undefined) {
             proxyUrl = this.config.proxy;
-        } else {
+        } else if (!noProxy(host,port)) {
             proxyUrl = process.env.HTTPS_PROXY || process.env.HTTP_PROXY;
         }
         if (proxyUrl) {
             agent = new HttpsProxyAgent(proxyUrl);
+        }
+
+        function noProxy(host, port) {
+            host = host.replace(/^\.*/,'.');
+            var noProxyString = process.env.NO_PROXY || process.env.no_proxy;
+            if (!noProxyString) {
+                return false;
+            }
+            if(noProxyString === '*') {
+                return true;
+            }
+            var noProxyList = noProxyString.split(',').map( function(s) {return s.split(':',2)} );
+            return noProxyList.some(function(noProxyEntry) {
+                var noProxyRegExp = new RegExp(noProxyEntry[0].replace(/^\.*/,'.').split('.').join('\\.')+'$');
+                return noProxyRegExp.test(host,'i') && (!noProxyEntry[1] || noProxyEntry[1]==port);
+            });
         }
         
         var ca = this.config.ca;


gr2m:
Hey there @goodefroi sorry that it took so long to get back to you! Is this pull request still relevant to you?

I have no experience myself in such a corporate environment like the one you describe. What is it you want to achieve with the `NO_PROXY` environment variable? What values could it have for which cases? We should definitely document `NO_PROXY` in the README if we will add the feature

goodefroi:
Hi @gr2m , yes, the pull request is still relevant. Imagine the following scenario: When accessing the Internet from inside the corporate network, all traffic shall be routed through a proxy. Therefore, `HTTP_PROXY` and `HTTPS_PROXY` are set to a specific proxy located inside the network that handles all of those outbound requests. Some services, however, might be located inside the network (and in our specific case, one of those services is an installation of GitHub Enterprise). These services are not accessible via proxy since the proxy only handles traffic to and from the WWW. Therefore, some internally-used domains are stored in the `NO_PROXY` environment variable (a simple string might look like this: `firstservice.mycompany.corp,github.mycompany.corp,additional.service.corp`) on corporate machines. This prompts applications to use the proxy specified in `HTTP_PROXY` and `HTTPS_PROXY` for all domains except those listed in `NO_PROXY`. 

The implementation submitted in the pull request mirrors the behavior of [request](https://github.com/request/request) in that an empty variable is ignored, `*` bypasses the proxy setup completely and a comma-separated list of domains like the one given above ignores proxy settings for all given domains and all their subdomains. If you still have any questions, be sure to let me know.

gr2m:
Thanks a lot for the clarification @goodefroi 👍 

I’m currently evaluating the code of `node-github` and try to come up with a way to make it work in browsers, too. That will require a few changes to the API and I want to make sure that your use case does not break when we move forward with that. For example, we will likely drop support for environment variables, all configuration will be needed to be passed to the `GitHub` constructor explicitly.

My understanding is that you are using multiple instances of `node-github` to connect to both:

1. https://api.github.com: all these requests go to through a proxy
2. your GitHub enterprise API, hosted inside your network: requests must not go through a proxy

The correct solution to me would be to differentiate the two instances by configuration that you pass into them. The current `GitHub` implementation already has a `proxy` configuration option that you can pass in: https://github.com/mikedeboer/node-github#example. Would it be a workable workaround for you not to use the `HTTP_PROXY` / `HTTPS_PROXY` environment variables at all, and instead pass in the `proxy` option explicitly to the instances it applies to?

goodefroi:
Thank you for your quick answer and your concerns about breaking my use case. Well, if you plan on ignoring environment variables altogether, there's not much sense in putting any further work into this pull request. My problem resulted from the inconsistent handling of environment variables (if you take `HTTP_PROXY` and `HTTPS_PROXY` into account, please don't ignore `NO_PROXY`). Dropping support for `HTTP_PROXY` and `HTTPS_PROXY` will actually solve all my issues without me having to do anything since I only need access to the internal GitHub Enterprise server. Other users might be in for a surprise though...
Thank you for your time, I think we can close this now.

gr2m:
Thanks @goodefroi! Things are still in the flux, I’ll leave your PR open as a reminder to make sure that we either remove handling of all environment variables in the new version, or we add handling for `NO_PROXY` as you suggest 👍

Issue #122, opened 2014-02-04
Tests seem to be failing?


davidascher:
```
davida@suki ~/s/node-github> mocha api/v3.0.0/authorizationTest.js 

  ․․․․․

  3 passing (8s)
  2 failing

  1) [authorization] should successfully execute GET /authorizations (getAll):
     Uncaught AssertionError: 41 == 2
      at /Users/davida/src/node-github/api/v3.0.0/authorizationTest.js:48:32
      at /Users/davida/src/node-github/api/v3.0.0/authorization.js:59:17
      at IncomingMessage.<anonymous> (/Users/davida/src/node-github/index.js:752:21)
      at IncomingMessage.EventEmitter.emit (events.js:117:20)
      at _stream_readable.js:910:16
      at process._tickCallback (node.js:415:13)

  2) [authorization] should successfully execute PATCH /authorizations/:id (update):
     Error: timeout of 2000ms exceeded
      at null.<anonymous> (/usr/local/lib/node_modules/mocha/lib/runnable.js:175:14)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
```

gr2m:
This issue is very outdated so I’m going to close it :) We are working on bringing back tests as part of the transition to the @octokit org: https://github.com/octokit/discussions/issues/7

PR #568, opened 2017-08-09
GitHub Integrations => GitHub Apps

bkeepers:
GitHub Integrations were renamed to GitHub Apps. The old endpoints will be [deprecated on Nov 22](https://developer.github.com/changes/2017-07-19-removal-of-integrations-routes/). This updates the routes to point to the new endpoints.

~~Note that this is a breaking change. It's likely possible to implement in a way that is backward compatible with deprecation warnings, but that doesn't seem to be a pattern that's been used on this library in the past.~~ Update: I added support for deprecating routes: 37c954d

I'll remove the `WIP` tag from the title once this has been tested and is ready for merge.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -547,28 +547,28 @@ declare namespace Github {
       tree: string;
       base_tree?: string;
     };
-  export type IntegrationsGetInstallationsParams =
+  export type AppsGetInstallationsParams =
     & Page
     & PerPage
     ;
-  export type IntegrationsCreateInstallationTokenParams =
+  export type AppsCreateInstallationTokenParams =
     & InstallationId
     & {
       user_id?: string;
     };
-  export type IntegrationsGetUserIdentityParams =
+  export type AppsGetUserIdentityParams =
     & {
       nonce?: string;
     };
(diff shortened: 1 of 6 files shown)


bkeepers:
I pushed some updates. Instead of replacing the old `integrations` routes, I update the URLs for them, marked them as deprecated, and then added the new `apps` routes.

zeke:
@bkeepers I'm still seeing lots of test failures. What do you think about disabling all the failing tests and just starting from scratch with the few that you've added here?

@mikedeboer do you have a Travis account? If so, would you be willing to enable it for this repo?

bkeepers:
> @bkeepers I'm still seeing lots of test failures. What do you think about disabling all the failing tests and just starting from scratch with the few that you've added here?

Yeah, I didn't do anything with the tests since my comment in #122. The tests added here are still just the generated ones and don't actually do anything.

jeffwilcox (approved):
LGTM

kaizensoze:
I'm waiting on updating everything else after doing a pass through the latest api doc so there's a potential major revision before creating a new one for this stuff, especially since the integration_id => app_id official deprecation doesn't happen until November 22, 2017.

PR #577, opened 2017-08-20
fix(types): Fix missing TypeScript interface generics

### LINKED ISSUE #576: TypeScript compilation error: "Generic type 'Array<T>' requires 1 type argument(s)"
Since version 9.3, the TypeScript compiler (in my case the latest 2.4.x) throws two errors regarding a missing generic for Arrays. In particular:

**First error:**
```
node_modules/github/lib/index.d.ts(1463,24): error TS2314: Generic type 'Array<T>' requires 1 type argument(s).
```
Here the line:
https://github.com/mikedeboer/node-github/blob/f34f59e2ea8d0c03c0a449e1c9f65de2fe2bb413/lib/index.d.ts#L1463

**Second error:**
```
node_modules/github/lib/index.d.ts(1471,24): error TS2314: Generic type 'Array<T>' requires 1 type argument(s).
```
Here the line:
https://github.com/mikedeboer/node-github/blob/f34f59e2ea8d0c03c0a449e1c9f65de2fe2bb413/lib/index.d.ts#L1471

**Solution:**

I assume using ```string[]``` or ```Array<string>``` in both cases should fix the issue.

itsdevdom:
**This PR fixes typings issues, in particular arrays without generics.** The fix itself was made in the ```generate.js``` file, and affects both the TypeScript and Flow typings.

This issue first occured with node-github version ```9.3```.

Closes #576.

### DIFF
--- a/lib/generateTypes.js
+++ b/lib/generateTypes.js
@@ -41,10 +41,12 @@ function replaceArrayTypes(type, name) {
         case "required_contexts":
         case "maintainers":
         case "reviewers":
+        case "team_reviewers":
         case "comments":
         case "labels":
         case "teams":
         case "users":
+        case "names":
         case "emails":
             if (type === "Array") {
                 return "string[]";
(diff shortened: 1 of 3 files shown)


kaizensoze:
Looks good. Thanks!

PR #584, opened 2017-09-01
remove obsolete .npmignore

gr2m:
Instead, white list files to be published

- If there is no .npmignore, npm publish is looking for a .gitignore. As both these files have the same value right now, we don’t need .npmignore: https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
- README.md and LICENSE are published implicitly and don’t need to be listed in `"files"`: https://docs.npmjs.com/files/package.json#files


kaizensoze:
Should tsconfig.json be in the files array if it's in the gitignore?

gr2m:
Good question Joe! I don’t know. I’m not using Typescript myself. I thought it would be a meta file used by IDEs to get information about the different methods, but looking at http://www.typescriptlang.org/docs/handbook/tsconfig-json.html it seems it’s only needed for development. I will amend my PR

kaizensoze:
Thanks!

PR #590, opened 2017-09-24
@octokit/fixtures^1.4.0

gr2m:
Tests will pass once https://github.com/node-nock/nock/pull/978 gets merged & relased


gr2m:
ugh sorry :(

PR #595, opened 2017-10-01
test: create file

gr2m:
adds integration test for `github.repos.createFile()`


gr2m:
I’ve run tests locally and they pass

PR #611, opened 2017-10-24
rename repo owner to octokit

gr2m:
in CONTRIBUTING.md and examples

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,10 @@
     {
       "name": "Joe Gallo",
       "email": "joe@brassafrax.com"
+    },
+    {
+      "name": "Gregor Martynus",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -15,6 +15,10 @@
     {
       "name": "Joe Gallo",
       "email": "joe@brassafrax.com"
+    },
+    {
+      "name": "Gregor Martynus",
+      "url": "https://github.com/gr2m"
```
gr2m, line 21: dunno if we should add contributors here, maybe we should remove contributors altogether from the package.json file? I think maintaining a list on github without bloating the package.json is a much better idea. any thoughts? 

I’ll get this merged but please let me know if you have any opinions

PR #623, opened 2017-10-25
remove custom Promise option

### LINKED ISSUE #622: remove custom Promise handling
All Node versions we currently support (Node 4, 6 and 8) have built-in Promise support, so we plan to remove the `Promise` options from the Constructor.

This like this will have no effect after v13.0.0

```js
const Client = require('./../lib/index')
const github = new Client({
  Promise: Promise
})
```

Please let me know if this causes any trouble for you, maybe we can find a workaround that we can add to the migration guide :) 

Related discussions are at #262, #258, #232

gr2m:
closes #622. This pull request will be merged into [milestone-v13 branch](https://github.com/octokit/node-github/tree/milestone-v13) to combine breaking changes into a single release

### BREAKING CHANGE

Passing a custom `Promise` option to the constructor is no longer supported

```js
const Client = require(./../lib/index)
const github = new Client({
  Promise: Promise
})
```

All suppertod Node versions (4, 6 and 8) have Promise support built in now. So migrating should be as simple as removing the `Promise` option altogether.

Everyone's reviews are welcome :) Let me know if you have questions


lib/index.js
```diff
@@ -81,7 +80,6 @@ var Client = module.exports = function (config) {
   config.headers = config.headers || {}
   this.config = config
   this.debug = Util.isTrue(config.debug)
-  this.Promise = config.Promise || config.promise || Promise
```
bkeepers, line 84: Is it worth logging a deprecation warning if `config.Promise` or `config.promise` is passed?

bkeepers (approved):
Are there reasons why someone might still want to use a custom promise library?

I'm fine with either way.

lib/index.js
gr2m: yes, good point, won't hurt

gr2m:
There are custom implementations that are compatible with the standard `Promise` object but have additional features on top. It is possible that people out there rely on these, but I think overall people migrate to just use the standard `Promise` now that it has been standardised and is available across all Node Versions and most browsers

Issue #624, opened 2017-10-25
Run examples as part of integration tests


gr2m:
this might help with #621

PR #626, opened 2017-10-25
adapt for mime@2

### LINKED ISSUE #625: v12.0.0 regression: mime.lookup has been renamed to mime.getType
I found out when running the [enterpriseUploadAsset.js example](https://github.com/octokit/node-github/blob/c961627ba2ff73f74eba72acf96779c08beb03c5/examples/enterpriseUploadAsset.js)

I'm looking into it, sorry if that caused you any trouble!

gr2m:
mime.lookup() is now mime.getType(). Closes #625.

I have a test locally that confirms the fix, the test is part of another upcoming PR for #624

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -848,7 +848,7 @@ var Client = module.exports = function (config) {
           callCallback(err)
         } else {
           headers['content-length'] = stat.size
-          headers['content-type'] = mime.lookup(msg.name)
+          headers['content-type'] = mime.getType(msg.name)
           httpSendRequest()
         }
       })


lib/index.js
```diff
@@ -848,7 +848,7 @@ var Client = module.exports = function (config) {
           callCallback(err)
         } else {
           headers['content-length'] = stat.size
-          headers['content-type'] = mime.lookup(msg.name)
+          headers['content-type'] = mime.getType(msg.name)
```
gr2m, line 851: see https://www.npmjs.com/package/mime#version-2-notes, sorry that I missed that

bkeepers (approved):

PR #631, opened 2017-10-26
replace custom extend method with Object.assign

### LINKED ISSUE #630: use Object.assign instead of a custom extend method

gr2m:
closes #630

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -742,7 +742,7 @@ var Client = module.exports = function (config) {
         headers[headerLC] = customHeaders[header]
       })
     }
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -742,7 +742,7 @@ var Client = module.exports = function (config) {
         headers[headerLC] = customHeaders[header]
       })
     }
-    addCustomHeaders(Util.extend(msg.headers || {}, this.config.headers))
+    addCustomHeaders(Object.assign(msg.headers || {}, this.config.headers))
```
gr2m, line 745: that was the only place that `Util.extend` was used 🤷

bkeepers (approved):

PR #632, opened 2017-10-26
remove Util methods

gr2m:
only Util.log left, I willll probably replace that with npm.im/debug in future

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -105,13 +105,7 @@ for (var status in statusCodes) {
 
   Util.inherits(error, exports.HttpError)
 
-  var className = toCamelCase(defaultMsg)
+  var className = defaultMsg.replace(/\s/g, '')
   exports[className] = error
   exports[status] = error
-}
-
-function toCamelCase (str) {
-  return str.toLowerCase().replace(/(?:(^.)|(\s+.))/g, function (match) {
-    return match.charAt(match.length - 1).toUpperCase()
-  })
 }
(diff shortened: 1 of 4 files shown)


lib/error.js
```diff
@@ -105,13 +105,7 @@ for (var status in statusCodes) {
 
   Util.inherits(error, exports.HttpError)
 
-  var className = toCamelCase(defaultMsg)
+  var className = defaultMsg.replace(/\s/g, '')
```
bkeepers, line 108: This looks like it will change functionality. What are the ramifications?

bkeepers (commented):
> only Util.log left, I willll probably replace that with npm.im/debug in future

I'd greatly appreciate if the logger could be made configurable. For example, Probot uses [bunyan](https://www.npmjs.com/browse/keyword/bunyan), which has cool features for logging and formatting objects so you can query logs by different properties.

lib/error.js
gr2m: In the context of this file, the `toCamelCase` method just turned messages like `Requested Range Not Satisfiable` into an Error name like `RequestedRangeNotSatisfiable`. It used its own toCamelCase method, not the Util.toCamelCase one

https://github.com/octokit/node-github/blob/fb0ad883f6f79007d41f6a94258c9b8a09fc2454/lib/error.js#L113-L117

I've double checked that, it's all good

gr2m:
@bkeepers 

> I'd greatly appreciate if the logger could be made configurable 

we have to take into account that bundle size for browser usage is a concern. It should be fairly simple to add custom logging with pre/post request hooks as needed, I'd keep the core one minimal

bkeepers:
> we have to take into account that bundle size for browser usage is a concern

Just allowing me to inject a logger object in the constructor that conforms to the `console` interface would be ideal.

Issue #635, opened 2017-10-26
browserify / webpack compatibility


gr2m:
This is a meta issue towards [browser support](https://github.com/octokit/node-github/issues/620#browser-support)

- [x]  #638 `github.repos.uploadAsset(options)`: replace `options.filePath` with `options.file`
- [x] make `node-github` work in the browser via [browserify](http://browserify.org/)
- [x] setup automated browser tests
- [x] measure bundle size to have a reference point
- [x] use browser native `fetch` as request module, add polyfill for node.js and mention [browser polyfill](https://github.com/github/fetch) for older browsers
- [x] add integation test to add webpack compatibility

related

- [x] #633: remove built-in netrc support

previous discussions

- #358 Fixed Browserify compatibility [for Node]

gr2m:
The latest `@octokit/rest` is already compatible with browserify & webpack.

I’m working on further optimizations, see my latest update at https://github.com/octokit/rest.js/issues/620#issuecomment-362715478

PR #637, opened 2017-10-26
move definitions out of routes.json

gr2m:
part of #635. Closes #358


gr2m:
@lolasc thanks for the thumbs up :) If you could use GitHub's review feature, that'd be great. It's a good way for us to see that you are looking at our PRs helping spotting oversights and that you are interested in getting involved ;)

bkeepers (approved):

PR #651, opened 2017-11-21
Fix link to promise example

vkentta:
* Remove outdated links
* Add new link to current example

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,8 +163,7 @@ github.authorization.create({
 
 ## Promises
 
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Kiitos, Ville! 👍

zeke:
TIL

image

I wonder if that's where "kudos" came from?

vkentta:
Kudos is actually also a Finnish word. Little bit different meaning though :smile:

zeke:
What's the difference?

vkentta:
https://fi.wikipedia.org/wiki/Kudos

PR #654, opened 2017-11-22
github.repos.uploadAsset(): adapt to specification

gr2m:
### Breaking change

The current implementation has the upload URL hard coded, while the specification says that it has to be received from the Release API endpoint:
https://developer.github.com/v3/repos/releases/#upload-a-release-asset

The new implementation also requires `contentType` as well as `contentLength`. The latter is not documented, but is required. The `filePath` option has been replaced with a `file` option which can be either a read stream, a buffer or a string.

Altogether, uploading an asset before worked like this:

```js
github.repos.getReleaseByTag({
  owner: 'octokit-fixture-org',
  repo: 'release-assets',
  tag: 'v1.0.0'
})

.then(result => {
  return github.repos.uploadAsset({
    owner: 'octokit-fixture-org',
    repo: 'release-assets',
    id: result.data.id,
    filePath: pathResolve(__dirname, 'test-upload.txt'),
    name: 'test-upload.txt',
    label: 'test'
  })
})
```

Now it looks like this

```js
github.repos.getReleaseByTag({
  owner: 'octokit-fixture-org',
  repo: 'release-assets',
  tag: 'v1.0.0'
})

.then(result => {
  return github.repos.uploadAsset({
    url: result.data.upload_url,
    file: 'Hello, world!\n',
    contentType: 'text/plain',
    contentLength: 14,
    name: 'test-upload.txt',
    label: 'test'
  })
})
```

If you prefer a higher-level release asset upload library, have a look at https://github.com/gr2m/octokit-release-asset-upload

### DIFF
--- a/lib/definitions.json
+++ b/lib/definitions.json
@@ -456,6 +456,27 @@
             "validation": "",
             "invalidmsg": "",
             "description": "Logins for Users to assign to this issue. NOTE: Only users with push access can set assignees for new issues. Assignees are silently dropped otherwise."
+        },
+        "url": {
+            "type": "String",
+            "required": true,
+            "validation": "",
+            "invalidmsg": "",
+            "description": "Dynamic URL for release asset uploads returned by the release’s API response."
+        },
+        "contentType": {
+            "type": "String",
+            "required": true,
+            "validation": "",
+            "invalidmsg": "",
+            "description": "The content type of a release asset upload."
+        },
+        "contentLength": {
+            "type": "Number",
+            "required": true,
+            "validation": "",
+            "invalidmsg": "",
+            "description": "Size of release asset upload in bytes."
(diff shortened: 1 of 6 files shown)


gr2m:
@bkeepers requested a review from you because of the idea of having the REST client APIs be generated from JSON. I had to extend the existing JSON for the edge cases that only apply to uploading assets, so this is just fyi. Happy to hear if you have any feedback of course, just didn’t want you to miss it :)

gr2m:
I’ll merge this in to finish up the 100% coverage, but let me know if you have any feedback, this will not get released immediately

bkeepers:
@gr2m thanks for the ping. This looks good to me. It's not ideal, but I know the upload API is an odd duck.

Issue #620, opened 2017-10-25
💌📯 Updates


gr2m:
If you are interested on updates, you can subscribe to this issue. I'll post weekly-ish comments with project updates. 

## Latest update

⬇ [Feb 20, 2020](https://github.com/octokit/rest.js/issues/620#issuecomment-589285596)

<a name="next-steps"></a>
## Next steps / Progress

We will keep releasing the package as `github` on npm for foreseeable time, before we eventually deprecate it and release `octokit` instead. We will give you an advanced notice and prepare a migration guide. We will also keep maintaining `github` for some time and release bug/security fixes, but will no longer add new features.

<a name="maintainability"></a>
### ✅  Maintainability 

<details>
  <summary>
    Released as  <a href="https://github.com/octokit/node-github/releases/tag/v13.0.0">v13.0.0</a>
  </summary>
In order to quickly accept contributions, we have to trust our test setup. Once a contribution is merged, the release should happen immediately to retain contributors

- [x] **Tests**. I’ve started implementing tests based on [our examples](https://github.com/octokit/node-github/tree/master/examples). We created a separate [Octokit Fixtures](https://github.com/octokit/fixtures) project which detects changes to the GitHub API using a daily cron job on travis. You can see the current status at  https://github.com/octokit/fixtures/issues/14, I’d love to take the opportunity to onboard new contributors, just [comment on the issue](https://github.com/octokit/fixtures/issues/14) if you are interested.
- [x] **Linting.** Thanks @Zeke! (https://github.com/octokit/node-github/pull/600)
- [x] #621 **Test coverage**. The goal is 💯
- [x] **Automate releases and dependency updates**. Using [semantic-release](https://github.com/semantic-release/semantic-release) & [Greenkeeper](https://greenkeeper.io/). Done via [#617](https://github.com/octokit/node-github/pull/617) & [#618](https://github.com/octokit/node-github/pull/618)
- [x] **Drop support for unmaintained Node versions**. We already just test Node 4, 6 and 8 on Travis. Done via [v12.0.0](https://github.com/octokit/node-github/releases/tag/v12.0.0)

  With minimal support of Node 4, we can
  - [x] #622 remove custom Promise handling
  - [x] #630 use Object.assign instead of a custom extend method
  
- [x] #666 **automate API DOCS generation** at http://octokit.github.io/node-github/
</details>

<a name="browser-support"></a>
### ✅ Browser support
<details>
  <summary>
[...]

PR #664, opened 2017-11-28
routes.json: add child team list route

nickvanw:
Hello 👋! 

I was looking do some work with the GitHub API with regards to child teams, and I noticed that this route (thus, the accompanying method) seemed to be missing! 

It's present in the hellcat preview definition [here](https://github.com/octokit/node-github/blob/1e07daf8f7774d02cdd2b3dd8a01eaebd9694f3d/lib/definitions.json#L533), so it should work with just these changes.

This is my first contribution to this project, so I'm not sure what the culture is around testing and examples? Let me know if there's anything else that needs to be done to get this merge-ready 😄.

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -2886,6 +2886,16 @@
             },
             "description": "List team members"
         },
+        "get-child-teams": {
+            "url": "/teams/:id/teams",
+            "method": "GET",
+            "params": {
+                "$id": null,
+                "$page": null,
+                "$per_page": null
+            },
+            "description": "List child teams"
+        },
         "get-team-membership": {
             "url": "/teams/:id/memberships/:username",
             "method": "GET",


gr2m:
Hello @nickvanw, thanks for the PR, it looks just right 👌 

You can follow all important updates of `node-github` at #620. We are nearly done with the "maintainability" milestone, after which I plan to take an inventory, because there are other methods currently missing. But no reason to make you wait, I’ll merge it in as is, it will get released automatically in am oment

gr2m:
released as [v12.1.0](https://github.com/octokit/node-github/releases/tag/v12.1.0)

nickvanw:
> You can follow all important updates of node-github at #620. We are nearly done with the "maintainability" milestone, after which I plan to take an inventory, because there are other methods currently missing.

Fantastic! Thanks for pointing me over there, and thanks for not making me wait 😁!

PR #667, opened 2017-12-01
Respect NO_PROXY environment variable

danielmoore:
This PR is addressing the same issue as #557, but delegates the logic to the well-tested [proxy-from-env](https://github.com/Rob--W/proxy-from-env) library.

I think this should be released as a patch that simply fixes an issue with the current support for environment variables.

I've also included a small fix to the tests. For some reason, `nope` on my MacOS High Sierra system resolves to one of my external IP addresses, which makes the test non-deterministic. I've changed it to attempt to connect to `127.0.0.1:8` which will reliably refuse a connection. Intriguingly, this is also a good test of the proxy handling.

closes #557

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var HttpsProxyAgent = require('https-proxy-agent')
+var getProxyForUrl = require('proxy-from-env').getProxyForUrl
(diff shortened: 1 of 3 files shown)


gr2m (approved):
I see no reason not to :) Thanks 👍

PR #671, opened 2017-12-09
fix: CLI debug instructions

paulmelnikow:
It seems the debug topic is always `node-github`, so the colon causes this not to match anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ github.authorization.create({
 
 ## DEBUG
 
-Set `DEBUG=node-github` for additional debug logs.
+Set `DEBUG=node-github*` for additional debug logs.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -157,7 +157,7 @@ github.authorization.create({
 
 ## DEBUG
 
-Set `DEBUG=node-github:*` for additional debug logs.
+Set `DEBUG=node-github` for additional debug logs.
```
gr2m: Oh good catch! Can we change it to `DEBUG=node-github*`, without the colon? That’ll make it future proof

paulmelnikow: Sure!

gr2m (approved):
cheers!

PR #672, opened 2017-12-22
For getPageLinks(link), when link is undefined or null it will throw …

MichaelTsengLZ:
…error.

This breaks at least two of my services. I think many other people have the same problem.


gr2m:
Thanks for the pull request @MichaelTsengLZ 👍 

I think the last time I looked at it, I couldn’t find a case where `link` would be undefined, sorry if I missed it. Could you add a test case to your pull request? 🙏

PR #673, opened 2017-12-22
make api reject 304s so caller can implement conditional requests

jsg2021:
I was attempting to implement conditional requests and hit an error in this library...

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -738,13 +738,13 @@ var Client = module.exports = function (config) {
           callCallback(err)
         })
         res.on('end', function () {
-          if (res.statusCode >= 301 && res.statusCode <= 307) {
+          if (res.statusCode != 304 && res.statusCode >= 301 && res.statusCode <= 307) {
             options.path = Url.parse(res.headers.location, true).path
             httpSendRequest()
             return
           }
 
-          if (res.statusCode >= 400 || res.statusCode < 10) {
+          if (res.statusCode == 304 || res.statusCode >= 400 || res.statusCode < 10) {
             callCallback(new error.HttpError(data, res.statusCode, res.headers))
           } else {
             res.data = data
--- a/lib/index.js
+++ b/lib/index.js
@@ -738,13 +738,13 @@ var Client = module.exports = function (config) {
           callCallback(err)
         })
         res.on('end', function () {
-          if (res.statusCode != 304 && res.statusCode >= 301 && res.statusCode <= 307) {
+          if (res.statusCode !== 304 && res.statusCode >= 301 && res.statusCode <= 307) {
             options.path = Url.parse(res.headers.location, true).path
             httpSendRequest()
             return
           }
 
-          if (res.statusCode == 304 || res.statusCode >= 400 || res.statusCode < 10) {
+          if (res.statusCode === 304 || res.statusCode >= 400 || res.statusCode < 10) {
             callCallback(new error.HttpError(data, res.statusCode, res.headers))
           } else {
             res.data = data
--- a/lib/error.js
+++ b/lib/error.js
@@ -45,6 +45,7 @@ Util.inherits(exports.HttpError, Error);
 }).call(exports.HttpError.prototype)
 
 var statusCodes = {
+  304: 'Not Modified', // See PR #673 (https://github.com/octokit/node-github/pull/673)
   400: 'Bad Request',
   401: 'Unauthorized',
   402: 'Payment Required',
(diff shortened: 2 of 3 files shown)


gr2m (changes requested):
Hey I’m sorry but I don’t quite understand your changes? `304` is not an error and hence shouldn’t be handled like one. Can you elaborate on the problem you are trying to solve? I can’t merge it like this.

jsg2021:
the location header is not present in a 304. So the first line change is
preventing a 304 from behaving like a redirect.

I thought about passing the the response as fulfilled payload, but it
doesn’t match the schema so, while it’s not an error, rejecting kinda made
since to me at first. What would you suggest? it does not have a body to
parse, the consumer would be responsible for catching and inspecting the
exception to return its cached response.

It seemed a bit much to expect this library to handle caching and 304s
directly.

jsg2021:
I’m writing a program to query my orgs repos and generate a report. It’s pretty request heavy so i need to add the conditional request headers so i don’t exhaust my api limits. 304s are currently falling because it’s trying to parse a url from a non existant location header.  I need to know explicitly that the response was a 304. While not an http error, it is an exception. when rased the caller can know nothing changed.

gr2m:
> the location header is not present in a 304

gotcha. Could you create a failing test for a 304 and we take it from there?

jsg2021:
sure thing. i’ll try to submit one tonight.

jsg2021:
Merry Christmas!

I added a test for 304. It will fail without my original changes. No matter how you expect it to resolve...because the original issue is the code is assuming all 301-307 codes are redirects. 304s do not have a location header, and the url parser blows up if the value passed to it is not a string.

My understanding & expectation is that 300s are still not 200's. They are still exceptions that must be dealt with. The library can transparently handle redirects... that's fine, but 304 and 308+ are still exceptions...not necessarily errors, but abort/short=circuit the request in some way. In my opinion, a rejection is appropriate.

gr2m:
Thank you @jsg2021, Merry Christmas to you, too 🎄 

I’ll check in with my fellow @octokit maintainers to see how they handle it in the kits for the other languages.

lib/index.js
```diff
@@ -738,13 +738,13 @@ var Client = module.exports = function (config) {
...
-          if (res.statusCode >= 301 && res.statusCode <= 307) {
+          if (res.statusCode !== 304 && res.statusCode >= 301 && res.statusCode <= 307) {
             options.path = Url.parse(res.headers.location, true).path
             httpSendRequest()
             return
           }
 
-          if (res.statusCode >= 400 || res.statusCode < 10) {
+          if (res.statusCode === 304 || res.statusCode >= 400 || res.statusCode < 10) {
             callCallback(new error.HttpError(data, res.statusCode, res.headers))
```
gr2m, line 748: Can you add a `304` status in https://github.com/octokit/node-github/blob/master/lib/error.js#L47-L48? Maybe add a comment to reference this pull request, as it looks a little confusing to have a single `304` besides all the 400+ error codes

gr2m (changes requested):

gr2m:
Okay I think I understand the context now. It’s correct that `node-github` is currently not handling `304` responses the way it should according to the [documentation on conditional requests](https://developer.github.com/v3/#conditional-requests).

I’d argue caching is within the scope of Octokit libraries, just like pagination and authentication, I’ll have to give it some more thought. I guess the main question is:

> Would you expect the caching to be handled by `node-github` automatically by keeping state of etags and sending the `If-None-Match` accordingly? 

Either way, the library does and will allow you to keep that state yourself, so it’s possible that a `304` response occurs without an internal state of the response. If the library would be able to keep state, it could resend the request and cache the response for future requests. In your case, I agree that the method call should respond with an error of some kind.

lib/index.js
jsg2021: sure thing, i’ll try to do it tomorrow

jsg2021:
Caching is a complicated realm. I would leave that to the consumers to decide on when/how/why to use caching. This library is more pure without that complicating things. 

With this tweak, opting into caching is pretty trivial, and in my opinion “good enough” :)

lib/index.js
jsg2021: i was thinking, along with more comments, I could make the 301-307 if statement predicate on the existence of a truthy location header instead of calling out 304... the in the exception case if statement switching it to be all status’s outside the 200-299 range... thoughts?

jsg2021: I've added the 304 to that file

gr2m:
Thanks @jsg2021 👍 A new release is on its way

PR #680, opened 2017-12-29
WIP – rename to @octokit/rest

### LINKED ISSUE #677: rename to `@octokit/rest`
As announced in https://github.com/octokit/node-github/issues/620#maintainability, it’s time to take the final step to make this the official Oktokit library for JavaScript.

- [x] #680 refactor code. Now with 100% test coverage, it is finally safe to do so
- [x] replace all mentions of `node-github` / `github` in code with `@octokit/rest`

moved remaining TODOs to the PR at #692 

Please comment here if you have any questions

gr2m:
closes #677

Don’t bother reviewing all commits, I think reviewing the final result and looking for obvious mistakes and for things that are unclear / could be made more clear, that would be great


gr2m:
I just realised this refactoring will break probot, because probot is using inheritance and the undocumented `github.handler` API to implement throttling.

The correct way would be a properly documented and tested plugin memchanism so probot can be upgraded, but I would like to merge the refactoring first. 

I wouldn’t consider it a breaking change as APIs like `.handler` or `.httpSend` have never been documented. But I wonder if other users of the library did similar things?

I’ll give give plugins some more thought

README.md
```diff
@@ -1,41 +1,18 @@
...
-## Example
-
-Get all followers for user "defunkt":
+## Usage
 
 <!-- HEADS UP: when changing the options for the constructor, make sure to also
      update the type definition templates in scripts/templates/* -->
 ```js
-var GitHubApi = require('github')
+const GitHubApi = require('github')
```
bkeepers: Will this still be `github`, or `require('@octokit/rest')`?

bkeepers (commented):
This is looking good!

README.md
gr2m: ha good catch!

README.md
```diff
@@ -206,6 +183,11 @@ For updates on endpoints under preview, see https://developer.github.com/changes
 | Timeline                | application/vnd.github.mockingbird-preview            |
 | Topics                  | application/vnd.github.mercy-preview+json             |
 
+## Credits
+
+`@octokit/rest` was originally created as `node-github` in 2012 by Mike de Boer from Cloud9 IDE, Inc.
```
zeke, line 188: Credits 👍

README.md
```diff
@@ -206,6 +183,11 @@ For updates on endpoints under preview, see https://developer.github.com/changes
...
+It was renamed adopted by GitHub in 2017
```
zeke: `renamed adopted` -> adopted and renamed

zeke:
What is the long-term plan for the `octokit` npm package name? Will that be a GraphQL client?

Once `@octokit/rest` lands on npm we should: 

```
npm deprecate github@"*" "The github package has been renamed to @octokit/rest"
```

gr2m:
> Once @octokit/rest lands on npm we should:

Yeah I have it as a todo in https://github.com/octokit/node-github/issues/677

> What is the long-term plan for the octokit npm package name? Will that be a GraphQL client?

I don’t know yet. We didn’t decide on what to do with GraphQL yet, I’m not sure we even need our own client at this point. 

`octokit` could be just a meta package that loads `@octokit/rest`, `@octkit/webhooks`, `@octokit/...` etc at once, the whole suite for the whole github platform. It’s a bigger discussion, but for this package we will go with `@octokit/rest`

gr2m:
also please folks don’t bother reviewing this PR in detail yet, I’m changing things quite a lot as I run into a few blockers with the first refactoring, I should have it ready by the end of this week

gr2m:
closing in favor of #692 which has the latest changes from master and cleaned up commits

PR #686, opened 2018-01-05
fix: correct typo in test/integration/lock-issue-test.js

francisfuzz:
Fixes https://github.com/octokit/node-github/issues/685 . :v:

`cc`: maintainers of `octokit/node-github` 👋


gr2m (approved):
great catch! Thanks for taking the time and sending a PR 🙏

PR #695, opened 2018-01-16
adds users.getInstallationRepos() to accept tree

aoberoi:
I'm currently getting the following error when trying to use `client.users.getInstallationRepos({ installation_id: id })`:

```
{"message":"If you would like to help us test the Integrations API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details.","documentation_url":"https://developer.github.com/v3"}
```

this patch adds the route to the accept tree so that the right `Accept` header is used.

### DIFF
--- a/lib/definitions.json
+++ b/lib/definitions.json
@@ -524,6 +524,7 @@
             "/apps/:app_slug",
             "/app/installations/:installation_id",
             "/user/installations",
+            "/user/installations/:installation_id/repositories",
             "/user/installations/:installation_id/repositories/:repository_id"
         ],
         "application/vnd.github.drax-preview+json": [


gr2m:
for the time being you can set the header explicitly

```js
client.users.getInstallationRepos({
  installation_id: id
  headers: {
    accept: 'application/vnd.github.machine-man-preview'
  }
})
```

We will remove setting these preview headers by default in a future version as the API behaviour can change during the API preview period

gr2m:
The `application/vnd.github.machine-man-preview` gets set implicitly if you use 

```js
client.authenticate({
  type: 'integration',
  token: '...'
})
```

So I think this should work now as expected. Let me know if it's still an issue for you

Issue #699, opened 2018-01-18
Missing TypeScript Types in v14.0.0, v14.0.2


fbartho:
I got excited to update to freshly released v14.0.0 to remove the rename-warning in our project, and went ahead and did the imports and ran the unit tests, and then realized there was a typescript compiler problem. Turns out the `index.d.ts` isn't being shipped with the new repo!

Context: https://github.com/danger/danger-js/issues/482

gr2m:
oh my I’m so very sorry, I think I know exactly what the problem is! Let me fix it right away

Thanks for testing it out that quickly and for reporting the problem :)

gr2m:
14.0.1 is on its way https://travis-ci.org/octokit/rest.js/builds/330145331

fbartho:
@gr2m build failed because ~merge closed the PR branch~ XD

gr2m:
oops wrong build sorry, this one: https://travis-ci.org/octokit/rest.js/builds/330145343

gr2m:
okay now something is broken with the release, I’ll have to look into it later tonight, sorry to keep you waiting

fbartho:
Woo!

fbartho:
Haha no worries & no pressure, this is not a pressing issue for me @gr2m

gr2m:
it should be fixed now :)

fbartho:
Bad news @gr2m -- it's still broken with 14.0.2 :( 

```
node_modules/@octokit/rest/index.d.ts(26,5): error TS1110: Type expected.
26     ;
```

Because of:
```typescript
declare namespace Github {

  export interface AnyResponse {
// …snipped...
  }

  export type WellKnownHeader =
    ;

  export interface EmptyParams {
  }
```

fbartho:
In commit: c6e5 index.js.flow.tpl had `#requestHeaders` deleted
https://github.com/octokit/rest.js/commit/c6e5#diff-6e934b6ee1e6925620b4e97cb684f4dfL3-L5

Looks like that needed to happen in the TypeScript [index.d.ts.tpl](https://github.com/octokit/rest.js/blob/25d0637e735cc317025c0d7a495cdb25267181ca/scripts/templates/index.d.ts.tpl#L26-L28) too! (Instead I think you get to delete the whole WellKnownHeaders type).

@gr2m -- please let me know if you'd like a new GitHub issue opened!

gr2m:
I'm very sorry for the trouble 😞 

[...]

PR #703, opened 2018-01-18
Updates the d.ts to have 0 errors

orta:
Fixes the d.ts noted in #699


gr2m (approved):
test it locally with `npx --package typescript tsc index.d.ts`. I’ll merge this without waiting for #704 so Typescript user are no longer blocked from using `@octokit/rest`

Thanks you 💐

PR #704, opened 2018-01-18
Runs the index.d.ts though the typescript compiler on CI

orta:
Runs the typescript compiler through the generated types on CI, so you will find out when it breaks - this should fail without a rebase on top of #703

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ stages:
 
 jobs:
   include:
-    # test stage: run Node in all maintained versions. Run coverage only in the last
+    # test stage: run Node in all maintained versions. Run unique checks only in the last
     - node_js: 4
     - node_js: 6
     - node_js: 8
@@ -25,6 +25,8 @@ jobs:
       script:
         - npm run test
         - npm run coverage:upload
+        - npm run build:ts
+        - npm run validate:dts
     # release stage: run semantic release & update the docs
     - stage: release
       node_js: lts/*
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "postbuild:docs": "apidoc -i doc/ -o doc/",
     "build:flow": "node scripts/generate-flow-types",
     "build:ts": "node scripts/generate-typescript-types",
+    "validate:dts": "npx --package typescript tsc index.d.ts",
     "deploy-docs": "gh-pages-with-token -d doc",
     "semantic-release": "semantic-release"
   },


package.json
```diff
@@ -70,6 +70,7 @@
     "postbuild:docs": "apidoc -i doc/ -o doc/",
     "build:flow": "node scripts/generate-flow-types",
     "build:ts": "node scripts/generate-typescript-types",
+    "validate:dts": "npx --package typescript tsc index.d.ts",
```
orta, line 73: For node 9, I think it's a safe bet that npx is installed.

orta:
Perfect, this is 👍

package.json
bkeepers: I'm not sure if there will be a speed difference, but to take advantage of module caching on Travis, it might be worth adding `npm install -g typescript` to the `install` script.

bkeepers (commented):
🎉  nice!

package.json
orta: I don't have too strong of an opinion TBH - I agree with where you're coming from, only difference is that this is a one-liner

.travis.yml
```diff
@@ -17,14 +17,16 @@ stages:
...
+    # test stage: run Node in all maintained versions. Run unique checks only in the last
     - node_js: 4
     - node_js: 6
     - node_js: 8
     - node_js: 9
       script:
         - npm run test
         - npm run coverage:upload
+        - npm run build:ts
+        - npm run validate:dts
```
gr2m, line 29: I think I would move `npm run validate:dts` into `postbuild:ts`, that way it will run in the release stage before the publish happens via semantic-release and it will be run automatically if people run `npm run npm run build:ts` locally.

Fixing a problem with the typescript definitions requires a new release, so it will be the right place to do it. What do you think?

package.json
gr2m: I would add `typescript` as dev dependency and the set the task to `"validate:ts": "typescript tsc index.d.ts"`. I would also add `"postbuild:ts": "npm run -s validate:ts"`.

Why `validate:dts` and not `validate:ts`? Is that a common convention? Should we rename `build:ts` to `build:dts`, too?

.travis.yml
orta: I was thinking this is more like a CI check, specifically because the last regression was due to making a change in the shared code for flow and TS and not propagating the changes to the typescript template. 

Now it would fail in that PR, instead of during a deploy later

package.json
orta: The created file is a `.d.ts` (a typescript definition file) so I figured I was just validating that the definition file passes the compiler's linter. Happy to add typescript as a dep etc 👍

orta: For simplicities sake I just made them both `ts`

gr2m:
I got a permission denied for updating the pull request, so I created a follow up PR here: https://github.com/octokit/rest.js/pull/704

orta:
rockin' thanks

PR #710, opened 2018-01-21
Fix typo in authentication example

not-an-aardvark:
This fixes `client_secert` to `client_secret` in the readme.


gr2m (approved):
thanks a lot @not-an-aardvark 💐

PR #713, opened 2018-01-22
fixes #711

### LINKED ISSUE #711: typings: getForRepo returns Promise<Github.AnyResponse> instead of a promise for an iterable
I used to be able to do

```
for (const issue of await octokit.issues.getForRepo({ owner, repo, state: 'open', labels: 'translation' })) {
...
}
```

but now I get

```
error TS2488: Type must have a '[Symbol.iterator]()' method that returns an iterator.
```

retorquere:
The typings were missing an iterator for methods that return arrays

fixes #711

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "build:flow": "node scripts/generate-flow-types",
     "build:ts": "node scripts/generate-typescript-types",
     "prevalidate:ts": "npm run -s build:ts",
-    "validate:ts": "tsc index.d.ts",
+    "validate:ts": "tsc --target es6 index.d.ts",
     "deploy-docs": "gh-pages-with-token -d doc",
     "semantic-release": "semantic-release"
   },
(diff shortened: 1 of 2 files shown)


gr2m:
looks like there is a syntax error now? https://travis-ci.org/octokit/rest.js/jobs/331701372#L2074

You can reproduce it locally by running `npm run validate:ts`

gr2m:
@orta could you please have a quick glance if this looks good?
https://github.com/octokit/rest.js/pull/713/files

orta:
@gr2m feel free to ping me any time 👍

orta:
@retorquere this PR makes sense in terms of the code in https://github.com/octokit/rest.js/issues/711, but TBH, I'm a bit surprised the old version worked - shouldn't you have to take the response out of `data` ?

retorquere:
I don't know, I've been delivering releases this way with 13.1 for a long while now, so it seemed to work  ¯\\\_(ツ)_/¯

orta:
Heh, well the d.ts is meant to describe the data - and if it works, it works

This is 👍 from me

gr2m (approved):

PR #719, opened 2018-01-25
Fix date parsing

### LINKED ISSUE #716: since parameter incorrectly formatted
I've been having a bug with requesting notifications while setting a `since` parameter. The following would not return a notification which came in at `2018-01-22T15:27:32Z`
```js
 octokit.activity.getNotifications({since: '2018-01-21T23:27:32Z' })
```
Upon researching, I found that this was the request getting sent to github 
> https://api.github.com/notifications?since=Mon%20Jan%2022%202018%2007%3A27%3A31%20GMT%2B0800%20(WITA)
>
> This corresponds to 2018-01-22

I don't think this is the correct format, [their api](https://developer.github.com/v3/activity/notifications/#list-your-notifications) expects an ISO string in the format `YYYY-MM-DDTHH:MM:SSZ`.

AFAICT, the error comes from [here](https://github.com/octokit/rest.js/blob/master/lib/plugins/endpoint-methods/validate.js#L59) and [here](https://github.com/octokit/rest.js/blob/master/lib/endpoint/add-query-parameters.js#L19).

The root of the problem is that the value this expression changes based on your system clock.
```js
encodeURIComponent(new Date('2018-01-21T23:27:31.000Z'))
```

I don't think the validate.js should cast dates strings to date objects. I think what it should do instead is try making a date object, and if that fails throw a `new errors.BadRequest` like other validations steps. If it succeeds, it should leave the datestring alone so it is passed to github. 

All this sound okay? I can make a PR for it

JuanCaicedo:
fixes #716

### DIFF
--- a/test/integration/params-validations-test.js
+++ b/test/integration/params-validations-test.js
@@ -127,4 +127,24 @@ describe('params validations', () => {
       due_on: new Date('2012-10-09T23:39:01Z')
     })
   })
+
+  it('Date is passed in correct format for notifications (#716)', () => {
+    const github = new GitHub({
+      host: 'notifications-test-host.com'
+    })
+
+    nock('https://notifications-test-host.com')
+      .get('/notifications')
+      .query(query => {
+        query.should.eql({
+          since: '2018-01-21T23:27:31.000Z'
+        })
+        return true
+      })
+      .reply(200, {})
+
+    return github.activity.getNotifications({
+      since: '2018-01-21T23:27:31.000Z'
+    })
+  })
 })
--- a/lib/plugins/endpoint-methods/validate.js
+++ b/lib/plugins/endpoint-methods/validate.js
@@ -55,14 +55,6 @@ function validate (endpointParams, options) {
       }
     }
 
-    if (expectedType === 'date') {
-      value = new Date(value)
-    }
-
-    if (parameter.mapTo) {
-
-    }
-
     set(options, parameter.mapTo || parameterName, value)
   })
 


test/integration/params-validations-test.js
```diff
@@ -127,4 +127,22 @@ describe('params validations', () => {
       due_on: new Date('2012-10-09T23:39:01Z')
     })
   })
+
+  it.skip('Date is passed in correct format for notifications', () => {
```
JuanCaicedo: I wasn't able to get the test suite to pass, so I figured I would add the tests, then get some feedback and figure out where to put it

test/integration/params-validations-test.js
```diff
@@ -127,4 +127,22 @@ describe('params validations', () => {
...
+    const github = new GitHub()
+
+    nock('https://api.github.com/notifications')
```
JuanCaicedo: Adding this mock causes other tests to fail. However, it doesn't seem like the changes should happen in [octokit/fixtures](https://github.com/octokit/fixtures) since we need to inspect the query parameter as part of the test. 

I'm not too familiar with nock to know why this is causing failures or what the proper mocking should be

test/integration/params-validations-test.js
JuanCaicedo: I added this test here since the bug stems from old code in the validations. I feel like the proper place would be in a file where other activity endpoints are being tested, but I'm not sure where that is

JuanCaicedo:
I've merged master and added another commit to fix the tests I had skipped. This is ready for review 😀

test/integration/params-validations-test.js
```diff
@@ -127,4 +127,24 @@ describe('params validations', () => {
       due_on: new Date('2012-10-09T23:39:01Z')
     })
   })
+
+  it('Date is passed in correct format for notifications', () => {
+    const github = new GitHub({
+      host: 'notifications-test-host.com'
```
JuanCaicedo: I saw this pattern used elsewhere of setting a host in order to locally mock the http requests. Not sure if this is the preferred technique, but I added it to get the tests working

gr2m: Thanks 👌 It is to prevent conflicts with [nock](https://github.com/node-nock/nock) mocks from other test files

test/integration/params-validations-test.js
```diff
@@ -127,4 +127,24 @@ describe('params validations', () => {
       due_on: new Date('2012-10-09T23:39:01Z')
     })
   })
+
+  it('Date is passed in correct format for notifications (#716)', () => {
```
gr2m, line 131: @JuanCaicedo I added the `(#716)` to the test description so folks can read up on the discussion behind this test

gr2m (approved):
Looks great :) Thank you 💐

PR #731, opened 2018-02-09
Several fixes for custom errors

pvdlg:
### Set the error code as a number for all errors

`Object.keys()` return and `Array` of `String` even if the object keys are not quoted (a `property` in a JS `Object` is always a `String`). So in the following snippet when calling `exports.HttpError.call(this, msg, status)`, `status` is a `String`.

https://github.com/octokit/rest.js/blob/47204015c20ee88926f8a17276cbbc236af3bc6d/lib/request/errors.js#L31-L36

But in the following one, `response.statusCode` is a `Number`. 

https://github.com/octokit/rest.js/blob/47204015c20ee88926f8a17276cbbc236af3bc6d/lib/request/request.js#L73

As a result some errors are thrown with `code` being a `String` and some other as a `Number`.

This PR guarantee the `code` is always a `Number` for each error.

### include stacktraces for each type of error

For some reason it seems the code intent is to include the stacktrace only for Error extending `HttpError` with a `code >= 500`:
https://github.com/octokit/rest.js/blob/47204015c20ee88926f8a17276cbbc236af3bc6d/lib/request/errors.js#L35

As previously mentioned `status` will be a `String` so comparing a `String` to a `Number` will always be `false` (`status >= 500` is always `false`).

Not including the stacktrace doesn't make much sense. It's useful for debugging and several library expect the `stack` property of the errors to be defined. Overall, swallowing the stacktrace seems to be a bad practice.

In addition `arguments.callee` is [deprecated and forbidden in strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee). The library works only because `if (status >= 500) { Error.captureStackTrace(this, arguments.callee) }` is never called due to `status` being a `String`.

This PR includes the stacktrace for each type of Error and use `Error.captureStackTrace(this, this.constructor)`.

### use ES6 classes to extend Error

Per [Node documentation](https://nodejs.org/docs/latest/api/util.html#util_util_inherits_constructor_superconstructor):

> Usage of util.inherits() is discouraged. Please use the ES6 class and extends keywords to get language level inheritance support. Also note that the two styles are semantically incompatible.

In addition the `name` property was not set. It's a common practice in JS (due to lack of type) to use the `name` to detect the type of Error. As it was not set `HttpError`  and it's subclass had the name `Error`.

This PR uses ES6 classes to extend Errors and properly set the error names.

### DIFF
--- a/lib/plugins/endpoint-methods/validate.js
+++ b/lib/plugins/endpoint-methods/validate.js
@@ -3,7 +3,7 @@
 module.exports = validate
 
 const set = require('lodash/set')
-const errors = require('../../request/errors')
+const HttpError = require('../../request/http-error')
 
 function validate (endpointParams, options) {
   Object.keys(endpointParams).forEach(parameterName => {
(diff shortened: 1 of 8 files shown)


pvdlg:
In addition the subclasses of `HttpError` should be removed as they don't seems to serve an purpose. They also make using the API harder as one would have to do something like this:
```js
try {

} catch(err) {
if (err.name === 'HttpError' || 'NotModified' || 'BadRequest' || 'NotFound' || 'InternalServerError' || 'GatewayTimeout') {

}
```

instead of:
```js
try {

} catch(err) {
if (err.name === 'HttpError') {

}
```

Also they don't respect the naming convention `xxxxError`.

Let me know if you want me to remove them.

gr2m:
Thanks for the great pull request 👍 

I agree that we should remove the HttpError subclasses, the differentiation using the status number makes more sense. It is still a relict of the previous `github` package. Removing them would be a breaking change though, while making sure the status is always a number is a fix. 

Maybe let's do that first and then do a follow up pull request removing the other subclasses that we can merge when we do a breaking release?

pvdlg:
> Maybe let's do that first and then do a follow up pull request removing the other subclasses that we can merge when we do a breaking release?

I don't know if that would be a breaking change though. Because currently the `name` is not set anyway. That would be a breaking change is one would do something like this:
```js
try {

} catch(err) {
if (err instanceof BadRequest) {

}
```
But that wouldn't be possible anyway as `BadRequest` is not exposed by the API.

gr2m:
You are right. Not that I doubted, just had to make double sure :) Yes, if you could remove the sub classes, that would be fabulous. Thank you :)

pvdlg:
Ok it's done, I added a fix commit to remove them

gr2m (approved):
fabulous work, thank you Pierre!

PR #741, opened 2018-02-17
options.agent for advanced request options in Node

gr2m:
Deprecates:

- `options.ca`
- `options.proxy`
- `options.family`
- `options.rejectUnauthorized`

### TODOs

- [x] add deprecated options to `deprecations-test.js` to make sure we don’t remove them unintentionally  
- [x] adopt typescript definitions

closes https://github.com/octokit/rest.js/pull/741


gr2m:
@orta @retorquere @philschatz This PR adds a new Constructor Option: `agent` and deprecates the ones above. `agent` is a [HTTP(s) Agent instance](https://nodejs.org/api/http.html#http_class_http_agent). 

I’d like to update the typescript definitions accordingly, I guess here:

https://github.com/octokit/rest.js/blob/968203a873e2c56736c700a32e46a794bdd5b602/scripts/templates/index.d.ts.tpl#L32-L44

Can you help me with that? Are there existing typescript definitions for Native Node classes like `http.Agent`?

philschatz:
@gr2m The TypeScript definitions for `'http'` are in [@types/node](https://www.npmjs.com/package/@types/node). Users of `rest.js` _may_ need to add it to their `dependencies` section in `package.json` but I am not sure.

And then, in `scripts/templates/index.d.ts.tpl` you should be able to just write

```typescript
import * as http from 'http'

export interface Options {
  // existing options here
  agent: http.Agent
}
```

gr2m:
<details>
<summary>
obsolete
</summary>
I tried the change

```
diff --git a/scripts/templates/index.d.ts.tpl b/scripts/templates/index.d.ts.tpl
index 4d25819..29ad27b 100644
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -41,6 +41,7 @@ declare namespace Github {
     requestMedia?: string;
     rejectUnauthorized?: boolean;
     family?: number;
+    agent?: http.Agent;
   }
 
   export interface AuthBasic  {
```

But I get this error

```
index.d.ts(44,13): error TS2503: Cannot find namespace 'http'.
```

I can look into it next week
</details>

gr2m:
I get this error: `index.d.ts(5,23): error TS2307: Cannot find module 'http'.` after 9c318db

gr2m:
Okay sorry it took me a while but now I get it, I have to install `@types/node` for the typescript validation to pass. Yeah looks like folks will need to add `@types/node` as dev dependencies, at least that’s what [angular](https://github.com/angular/angular/blob/a4032296cc1647cd2bac36d63304fadf60fea9d8/package.json#L46), [vue](https://github.com/vuejs/vue/blob/cd334070f3b82d3f5892c4999cc290ccd4f56fd8/package.json#L69) and [rxjs](https://github.com/ReactiveX/rxjs/blob/2f9efbfa1f2c9c9c0d08598cda287b99b3f8b87c/package.json#L173) are doing, and these are only the first I checked

PR #748, opened 2018-02-22
fix: correct parameters for pull request creation

alexander-fenster:
The current implementation does not work because it does not pass a `title` which is required. This quick fix makes the client behave as described in the [API documentation](https://developer.github.com/v3/pulls/#create-a-pull-request).

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -4998,12 +4998,10 @@
 	},
 	"body": {
 	  "type": "string",
-	  "required": false,
 	  "description": "The contents of the pull request."
 	},
 	"maintainer_can_modify": {
 	  "type": "boolean",
-	  "required": false,
 	  "description": "Indicates whether maintainers can modify the pull request."
 	}
       },
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -4990,7 +4990,22 @@
           "type": "string",
           "required": true,
           "description": "The branch (or git ref) you want your changes pulled into. This should be an existing branch on the current repository. You cannot submit a pull request to one repo that requests a merge to a base of another repo."
-        }
+        },
+	"title": {
+	  "type": "string",
+	  "required": true,
+	  "description": "The title of the pull request."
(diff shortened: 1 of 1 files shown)


lib/routes.json
```diff
@@ -4990,7 +4990,22 @@
...
-        }
+        },
+	"title": {
+	  "type": "string",
+	  "required": true,
+	  "description": "The title of the pull request."
+	},
+	"body": {
+	  "type": "string",
+	  "required": false,
```
gr2m: just leave out `required` unless it's true

lib/routes.json
```diff
@@ -4990,7 +4990,22 @@
...
+	  "description": "The contents of the pull request."
+	},
+	"maintainer_can_modify": {
+	  "type": "boolean",
+	  "required": false,
```
gr2m: same here just leave out `required` unless it's true

gr2m:
@alexander-fenster I was not able to reproduce the problem. `octokit` is passing the title property today as shown in this example: https://runkit.com/gr2m/5a8f640e1c7d85001208450d

Am I missing something?

alexander-fenster:
@gr2m the problem is that the following request
```javascript
octokit.pullRequests.create({
  owner: 'owner',
  repo: 'repo',
  head: 'head',
  base: 'base',
// no title
})
```
is in fact invalid (GitHub API will return failure because of no title), but neither octokit fails validation in this case, nor the documentation at https://octokit.github.io/rest.js/#api-PullRequests-create mentions that `title` is required.

lib/routes.json
alexander-fenster: removed this line

lib/routes.json
alexander-fenster: removed this line

gr2m:
Got it, thanks!

gr2m (approved):
thanks @alexander-fenster

PR #764, opened 2018-02-25
use node-fetch

### LINKED ISSUE #742: migrate to node-fetch
Instead of having two different request methods for Browsers and Node 

1. Node: [/lib/request/request.js](https://github.com/octokit/rest.js/blob/968203a873e2c56736c700a32e46a794bdd5b602/lib/request/request.js)
2. Browser: [/lib/request/request-browser.js](https://github.com/octokit/rest.js/blob/968203a873e2c56736c700a32e46a794bdd5b602/lib/request/request-browser.js)

We could migrate to [node-fetch](https://github.com/bitinn/node-fetch) instead. They released 2.0.0 which seems solid and I see more and more libraries migrate to it.

gr2m:
### TODOs

- ~~make `ocotkit.request` return response options (https://github.com/octokit/rest.js/issues/742#issuecomment-368351394)~~ https://github.com/octokit/rest.js/issues/777
- ~~make it simple to get custom response format for endpoint methods (https://github.com/octokit/rest.js/issues/742#issuecomment-368351394)~~ https://github.com/octokit/rest.js/issues/778
- ~~Figure out debug logging in browser, `DEBUG=octokit:rest*` is not working for Browsers :) maybe allow to pass a `log` option that can either be a function or a `console`-compatible object with `{debug, info, warn, error}` properties. It could default to the `debug` module in Node and an no-op function in the browser~~ #779
- ~~Depends on https://github.com/bitinn/node-fetch/pull/408 for the `github.repos.uploadAsset as ArrayBuffer` test to pass~~ #780
- ~~Depends on https://github.com/bitinn/node-fetch/pull/409 for correct 307 redirect handling~~ #780
- [x] test with popular dependents with good test suites
  - [x] danger-js
  - [x] greenkeeper
  - [x] probot
  - [x] release-it
  - [x] semantic-release-github

fixes #742


gr2m:
Removing the custom errors means to use [custom responses](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response). And these don’t work well with callbacks, so I think I will remove support for callbacks in the endpoint methods

Issue #774, opened 2018-03-01
webpack compatibility for serverless framework -> lambda


wyvern8:
Hi @gr2m , i have made some progress on the issue reported here https://github.com/octokit/rest.js/commit/d191dfc7096df038d767f2000d11caa81b719375#commitcomment-27295686 

There are two places causing issues with webpack.

### 1. is-array-buffer module.
webpack does not seem to like 'export default', although in this link they could use it by hoisting function first.  https://github.com/webpack/webpack/issues/2399

due to the package.json module field of is-array-buffer, the file picked up is **is-array-buffer.esm.js**.

this file uses es6: 
```
export default isArrayBuffer;
```
.. however if i change it to contain the same as 'is-array-buffer.common.js':
```
module.exports = isArrayBuffer;
```
..it works with webpack.  given this module is about 5 lines of code, it might be simplest to incorporate n octokit with attribution?  https://github.com/fengyuanchen/is-array-buffer/blob/master/dist/is-array-buffer.common.js 

### 2. webpack does not like require calls containing expressions.
https://github.com/webpack/webpack/issues/2413

in octokit get-request-agent.js

.. changing this:
```
const Agent = require(options.protocol || DEFAULTS.protocol).Agent
```
..to this..
```
let Agent;
if (options.protocol && options.protocol === 'http') {
	Agent = require('http').Agent;
} else {
	Agent = require('https').Agent
}	    
```
..works, but you might want to handle it differently.

**So, would you like a pull request with the above 2 items, or prefer to handle this yourself?**

i'm pretty keen to continue using webpack for our lambdas, as it bring size down from 64Mb to 3Mb each. fyi @Neko-Design

gr2m:
Thanks @wyvern8, could you send two separate pull requests please?

Just a quick tip, you past a permalink URL into a comment, it will show the actual code and referencing the right file, so it’s easier for others to understand where to code is you are referencing

![image](https://user-images.githubusercontent.com/39992/36867015-bfb29be8-1d48-11e8-90e9-8f8474e6a777.png)

e.g. `https://github.com/octokit/rest.js/blob/57c20ed25ca6da640571db0e9cea9dfab11597b3/lib/get-request-agent.js#L20` will result in:

https://github.com/octokit/rest.js/blob/57c20ed25ca6da640571db0e9cea9dfab11597b3/lib/get-request-agent.js#L20

[...]

PR #775, opened 2018-03-02
fix(request): replace is-array-buffer module with inline equivalent

wyvern8:
is-array-buffer causes issues with webpack

addresses #774

### DIFF
--- a/lib/request/is-array-buffer.js
+++ b/lib/request/is-array-buffer.js
@@ -1,0 +1,32 @@
+/*!
+ * isArrayBuffer v1.0.0
+ * https://github.com/fengyuanchen/is-array-buffer
+ *
+ * Copyright (c) 2015-2017 Chen Fengyuan
+ * Released under the MIT license
+ *
+ * Date: 2017-07-26T11:00:44.931Z
+ */
+
+'use strict'
+
+// temporary workaround for https://github.com/octokit/rest.js/issues/774
+
+const hasArrayBuffer = typeof ArrayBuffer === 'function'
+const toString = Object.prototype.toString
+
+/**
+ * Check if the given value is an ArrayBuffer.
+ * @param {*} value - The value to check.
+ * @returns {boolean} Returns `true` if the given is an ArrayBuffer, else `false`.
+ * @example
+ * isArrayBuffer(new ArrayBuffer())
+ * // => true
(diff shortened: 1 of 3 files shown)


lib/request/request.js
```diff
@@ -22,6 +21,22 @@ function httpRequest (origRequestOptions) {
...
+  * Released under the MIT license
+  *
+  * Date: 2017-07-26T11:00:44.931Z
+  */
+  function isArrayBuffer (value) {
+    var hasArrayBuffer = typeof ArrayBuffer === 'function'
+    var toString = Object.prototype.toString
+    return hasArrayBuffer && (value instanceof ArrayBuffer || toString.call(value) === '[object ArrayBuffer]')
+  }
+  // --- end isArrayBuffer
```
gr2m: could you please move it into ` lib/request/is-array-buffer.js`, mention that it’s a temporary workaround and reference https://github.com/octokit/rest.js/issues/774

.gitignore
```diff
@@ -19,3 +19,8 @@ dist/
 
 # cypress
 cypress/
+
+# intellij
+.idea/
+*.iml
+
```
gr2m: could you please remove this unrelated change?

You can set global gitignore rules in `~/.gitignore` so you don’t need to set it per project :)

wyvern8: yeah good point - want to leave it there for others, or remove?

lib/request/is-array-buffer.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Date: 2017-07-26T11:00:44.931Z
+ */
+
+'use strict'
+
+// temporary workaround for https://github.com/octokit/rest.js/issues/774
+
+var hasArrayBuffer = typeof ArrayBuffer === 'function'
+var toString = Object.prototype.toString
```
gr2m: can you use `const` please?

.gitignore
gr2m: please remove, thank you!

wyvern8: done

gr2m (dismissed):
thanks :)

gr2m:
I’ll merge & release it together with #776

gr2m (approved):

PR #776, opened 2018-03-02
Fix/webpack require expression

wyvern8:
addresses #774

### DIFF
--- a/lib/get-request-agent.js
+++ b/lib/get-request-agent.js
@@ -1,7 +1,8 @@
 module.exports = getRequestAgent
 
 const pick = require('lodash/pick')
-const DEFAULTS = require('./defaults')
+const HttpAgent = require('http').Agent
+const HttpsAgent = require('https').Agent
 
 function getRequestAgent (options) {
   if (options.agent) {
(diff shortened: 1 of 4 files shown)


lib/get-request-agent.js
```diff
@@ -17,7 +16,13 @@ function getRequestAgent (options) {
...
     console.warn(`options.${option} is deprecated. Use "options.agent" instead`)
   })
-  const Agent = require(options.protocol || DEFAULTS.protocol).Agent
+
+  var Agent
+  if (options.protocol && options.protocol === 'http') {
+    Agent = require('http').Agent
+  } else {
+    Agent = require('https').Agent
+  }
```
gr2m: I think style-wise I would prefer to just load both on top

```js
const HttpAgent = require('http').Agent
const HttpsAgent = require('https').Agent
```

or 

```js
const AGENTS = {
  http: require('http').Agent
  https: require('https').Agent
}
```

Do you see any reason not to do it that way?

gr2m (dismissed):
great, thanks :)

wyvern8:
thanks @gr2m !

PR #783, opened 2018-03-04
agent & deprecated {proxy, ca, rejectUnauthorized} options

### LINKED ISSUE #770: Proxy is not used
I have just moved from `github` to `@octokit/rest.js` (thanks for the How to Description). Unfortunately since then the requests does not work anymore at our company, because we have an proxy server where all requests have to be redirected to. My current setup is pretty easy: 

```js
const proxy = process.env.https_proxy
const GitHubApi = require('@octokit/rest')
const githubApiOptions = {
  proxy: proxy
}
const github = new GitHubApi(githubApiOptions) 
```

As your moving guide said I'm currently using `@octokit/rest` at version `14.0.9`. Before submitting this issue I have tried to find a similar one and found that with version `15.1.0`, you have changed the handling for the proxy configuration. Due to this line https://github.com/octokit/rest.js/blob/57c20ed25ca6da640571db0e9cea9dfab11597b3/lib/get-request-agent.js#L22

I have thought okay, so we will just to update to the current version `15.1.2` of `@octokit/rest`. But the request still does not go through the proxy. So I have changed my code to: 

```js
const GitHubApi = require('@octokit/rest')
const proxy = process.env.https_proxy
const https = require('https')
const githubApiOptions = {
  agent: https.Agent({
     proxy: proxy
  })
}
const github = new GitHubApi(githubApiOptions) 
``` 

What is more or less that the current version doing (without deprecation warnings). But still requests does not run through the proxy server.  So I have started to checkout the `https.Agent` documentation and found nothing about the `proxy` option attribute.  Which means that this won't work anyway. To fix this for my case I have made use of [https-proxy-agent](https://www.npmjs.com/package/https-proxy-agent) module and changed my code to:

```js
const GitHubApi = require('@octokit/rest')
const proxy = process.env.https_proxy
const HttpsProxyAgent = require('https-proxy-agent')

const githubApiOptions = {
  agent: HttpsProxyAgent(proxy)
}

const github = new GitHubApi(githubApiOptions)
```

This is working for my case, but i'm not sure how to generally fix this.

### LINKED ISSUE #781: Add test for "proxy" client option
follow up for https://github.com/octokit/rest.js/issues/770. Compare [test/integration/ssl/certificate-test.js](https://github.com/octokit/rest.js/blob/a3693becdc6b5ba41fc6d0f6f560ab9bb4e96d82/test/integration/ssl/certificate-test.js)

- [ ] add test with http proxy and remove the `istanbul ignore if` statement from  https://github.com/octokit/rest.js/blob/a3693becdc6b5ba41fc6d0f6f560ab9bb4e96d82/lib/get-request-agent.js#L22-L28

gr2m:
closes #781, closes #770


gr2m:
I tried to add tests for the `family` option, too, but to avail. It seems to work out of the box without the `family` option and the tests fail on Travis because of lack of IPv6 support, so it’s not work the effort. Here’s the test for reference

```js
const http = require('http')
const Octokit = require('../../..')

require('../../mocha-node-setup')

describe('IPv6', () => {
  let server
  before((done) => {
    server = http.createServer(function (request, response) {
      response.writeHead(200)
      response.write('ok')
      response.end()

      expect(request.headers.host).to.equal('[::1]:' + server.address().port)
    })

    server.listen(done)
  })

  // unless I’m missing something, IPv6 should just work (tm)
  // no matter if "family" option is set or not. Please correct me if I’m wrong
  it('host = [::1], without family option', () => {
    const octokit = new Octokit({
      protocol: 'http',
      host: '[::1]',
      port: server.address().port
    })

    return octokit.repos.get({
      owner: 'octokit',
      repo: 'rest.js'
    })
  })

  it('options.family', () => {
    const octokit = new Octokit({
      protocol: 'http',
      host: '[::1]',
      port: server.address().port,
      family: 6
    })

    return octokit.repos.get({
      owner: 'octokit',
      repo: 'rest.js'
    })
  })

  it('options.agent = new Agent({family: 6})', () => {
    const octokit = new Octokit({
      protocol: 'http',
      host: '[::1]',
      port: server.address().port,
      agent: new http.Agent({family: 6})
    })

    return octokit.repos.get({
      owner: 'octokit',
      repo: 'rest.js'
    })
  })

  after(done => server.close(done))
})
```

PR #802, opened 2018-03-08
deprecations and default baseUrl added

aryannarora:
@gr2m , I'd have to learn testing a bit, to update tests for these additions. 😅  And I couldn't find `options.requestmedia` anywhere in this file.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -1,10 +1,7 @@
 module.exports = {
   agent: undefined, // https://nodejs.org/api/https.html#https_class_https_agent
   headers: {},
-  host: 'api.github.com',
-  pathPrefix: '', // for GitHub Enterprise
-  port: undefined,
-  protocol: 'https',
+  baseUrl: 'https://api.github.com',
   requestMedia: 'application/vnd.github.v3+json',
   timeout: 0
 }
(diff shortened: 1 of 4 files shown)


gr2m:
you can use this test as an example:

https://github.com/octokit/rest.js/blob/bf2f550557bfed7499ef197d22c4c58f71754f02/test/integration/smoke-test.js#L12-L24

You can run the tests locally with `npm test`, does that work for you?

aryannarora:
@gr2m  Yes, i can. Actually my network was taking so long to download those npm dependencies so i did it here. I'll do it locally from next time 😅

aryannarora:
@gr2m Coverage test is showing uncovered lines in `lib/plugins/pagination/` . 😕 
and does the travis-ci needs new tests to be written for the added code to pass?

gr2m:
> Coverage test is showing uncovered lines in lib/plugins/pagination/ . 😕

You can run `npm run coverage` to see what lines are not covered locally, that’s usually very helpful. In that case the problem is probably that some of the tests fail, so they don’t get to the lines they would cover otherwise

> and does the travis-ci needs new tests to be written for the added code to pass?

It looks like existing tests broke. So something is not quite right yet

aryannarora:
@gr2m Could you please review the tests i wrote for deprecations? 😅 
Also, i tried running tests locally without altering any file and that still failed. 😕  I thought my code was breaking it. 😆

aryannarora:
Closing this PR as this branch has some unknown issues. Opened a new one #807.

PR #803, opened 2018-03-09
Ensure constructor options are optional (in TypeScript)

philschatz:
This was overlooked in #741 .

I was not sure if tests should be included but added one since `typescript` is already included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "generate-bundle-report": "webpack-bundle-analyzer dist/bundle-stats.json --mode=static --no-open --report dist/bundle-report.html",
     "prevalidate:ts": "npm run -s build:ts",
     "validate:ts": "tsc --target es6 index.d.ts",
-    "postvalidate:ts": "npm run-script build:ts && tsc test/typescript-validate.ts",
+    "postvalidate:ts": "tsc --noEmit test/typescript-validate.ts",
     "deploy-docs": "gh-pages-with-token -d doc",
     "semantic-release": "semantic-release",
     "start-fixtures-server": "octokit-fixtures-server"
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "coverage:upload": "nyc report --reporter=text-lcov | coveralls",
     "pretest": "standard && standard-markdown",
     "test": "nyc mocha test/mocha-node-setup.js 'test/**/*-test.js'",
-    "posttest": "npm run-script build:ts && tsc test/typescript-validate.ts ",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -86,6 +86,7 @@
     "coverage:upload": "nyc report --reporter=text-lcov | coveralls",
     "pretest": "standard && standard-markdown",
     "test": "nyc mocha test/mocha-node-setup.js 'test/**/*-test.js'",
+    "posttest": "npm run-script build:ts && tsc test/typescript-validate.ts ",
```
gr2m: we have `validate:ts` which we don’t run with each `npm test`. Maybe better to move do 

```
 +    "postvalidate:ts": "tsc test/typescript-validate.ts", 
```

Instead?

Also tiny nitpick: trailing space at `.ts "`

test/typescript-validate.ts
```diff
@@ -0,0 +1,15 @@
+import * as Octokit from '../'
+
+// This code is NOT executed.
+
+// Test the TypeScript definition.
+export default async function() {
+  new Octokit() // Check empty constructor
+  const octo = new Octokit({}) // Check that all arguments are optional
+  const repo = await octo.repos.get({owner: 'octokit', repo: 'rest.js'})
```
gr2m: I would like the tests to be able to run when offline. Could you add mocking with nock, or get the instance like we do in the scenario tests, e.g.

https://github.com/octokit/rest.js/blob/bf2f550557bfed7499ef197d22c4c58f71754f02/test/scenarios/get-repository-test.js#L8-L14

philschatz: This code was not intended to be run (it does not match `'test/**/*-test.js'`). 

The "test" is just that `tsc` does not complain when compiling the `.ts` file; it would complain if, for example, you edited the file to say `repo.meta.linkThisFieldIsInvalid`.

Is there a better place for this file to make it clear that the generated `.js` file is not actually executed?

gr2m: ah okay that makes sense. Maybe an explanatory comment in the `test/typescript-validate.ts` would help? It is all kind of confusing ... do you know if is there a command option for `tsc` that does not create a js file, only checks for compilation errors?

philschatz: I included a small comment at the top of the file but it seems like it was overlooked. In the most recent commit I made it a bit bigger 😄

gr2m: I was very tired when I checked that. New comment is great, thank you :)

gr2m (approved):
thanks Phil 💐

PR #807, opened 2018-03-09
Deprecations removed and tests added

aryannarora:
@gr2m Please review this. 😄


gr2m:
looks good so far! Still need to update README, defaults and setting of `clientDefaults.baseUrl` from `options.baseUrl` if it’s passed

aryannarora:
@gr2m In `lib/defaults.js`, should i replace `headers`, `port`, `pathPrefix`, `host` by `baseUrl` ?

I just tried it and i'm getting crazy amount of errors while testing. 😕 
Also, coverage must have decreased because of an added if statement.

gr2m:
> @gr2m In `lib/defaults.js`, should i replace `headers`, `port`, `pathPrefix`, `host` by `baseUrl` ?

yes :) 

> I just tried it and i'm getting crazy amount of errors while testing. 😕 

Hey I know you don’t mean to, but using "crazy" can make people uncomfortable who live with mental illness, so we try to avoid it. You can google for "crazy ableist" if you would like to learn more, there are other terms to avoid, too. It helps to keep communities welcoming to everyone.

I’ll have a look at the code in a moment and report back :)

gr2m:
Do you mean the deprecation warnings such as 

image

These are just the warnings that you implemented :) For most of the tests, the Octokit instance is created in `test/util.js#fixtureToInstace`

https://github.com/octokit/rest.js/blob/2962dcd72146590518f78b42ba29e00d2fdbca6a/test/util.js#L35-L39

There are lots of other places where we set one of the now deprecated options, could you do a search across all files for `host:` etc and replace them with the new `baseUrl` option?

aryannarora:
@gr2m I'm sorry, my bad, i'll surely avoid such words from now. 😃 
No, i was not talking about deprecated errors. These are some of the errors i got.



> error3

aryannarora:
@gr2m  Yes sure. I'll look for those files. :)
And the reason why i got these errors are those files only.
I replaced `host` with `baseUrl` in one of the files and corresponding tests passed.
Thanks for the lead. I'll update them all right away. 😄

lib/parse-client-options.js
```diff
@@ -38,18 +55,11 @@ function parseOptions (userOptions) {
...
-
-  if (options.port) {
-    clientDefaults.baseUrl += `:${options.port}`
-  }
-
-  // Check if a prefix is passed in the options and strip any leading or trailing slashes from it.
-  if (options.pathPrefix) {
-    clientDefaults.baseUrl += '/' + options.pathPrefix.replace(/(^[/]+|[/]+$)/g, '')
+  if (options.baseUrl) {
+    clientDefaults.baseUrl = `${options.baseUrl}`
```
gr2m: no need to wrap it in `${}`, just `clientDefaults.baseUrl = options.baseUrl` should work.

The previous implementation you had before looked good, why did you remove it? Now we won’t support the `protocol`, `host`, `port` and `pathPrefix` argument any longer. But we still want to support them so apps don’t break, they will only log the deprecation warning.

aryannarora: @gr2m I must have done that by mistake. I was playing around with some things. 😅 
Also, i'm having some trouble to pass some tests that are requiring `test/util.js` file. I believe the error is coming from its `fixtureToInstance()` function . Could you help me solving it?  😄 
Locally, if i run the tests it give the following error.

> error4

test/util.js
```diff
@@ -33,9 +33,7 @@ function fixtureToInstace ({url}, options) {
   url = parseUrl(url)
 
   return new GitHub(merge(options, {
-    host: url.host,
-    protocol: url.protocol.replace(/:$/, ''),
-    pathPrefix: url.path
+    baseUrl: url.protocol.replace(/:$/, '') + url.host + '/' + url.path
```
gr2m: @aryannarora very close! With the new `baseUrl` parameter, we don’t need to parse `url` option any more, instead we can pass it directly as `baseUrl`. We only parsed the URL because we had to pass `host`, `protocol` and `pathPrefix` separately before

aryannarora: @gr2m This whole time i was thinking that the response is in such a way that we have to parse it.
Thanks for this clarification. :)

aryannarora:
@gr2m One test is still failing. Here it is.

> err5

gr2m:
@aryannarora wanna try to figure this one out yourself? The error is about the `proxy` option which is being taken care of in `lib/get-request-agent.js`, will be likely something in there :)

aryannarora:
@gr2m I didn't quite understand what to look for in `lib/get-request-agent.js` . 😓 
I read the entire file but couldn't deduce the reason for that error. I'll go try again.

gr2m:
no worries :) we check for `options.protocol` twice in the file

https://github.com/octokit/rest.js/blob/2962dcd72146590518f78b42ba29e00d2fdbca6a/lib/get-request-agent.js#L36-L38

https://github.com/octokit/rest.js/blob/2962dcd72146590518f78b42ba29e00d2fdbca6a/lib/get-request-agent.js#L44-L46

But now there is no more `options.protocol`, we have `options.baseUrl` now. So to get the protocol we’ll have to do something like this

```js
const protocol = urlParse(options.baseUrl).protocol
```

and then check for that instead of `options.protocol`. Does that make sense?

aryannarora:
@gr2m Everything becomes so simple when you explain it. 😄

gr2m:
I know how you feel. But you got one bug left there, that I run into myself so many times my brain sees it immediately now. Open the Node REPL and run `url.parse('http://example.com').protocol` to see what I mean :)

aryannarora:
@gr2m You are making it so much fun for me. Thank you. 😄 
I feel so good now. That error was right in front of me, i even consoled it earlier but subconsciously my mind just ignored that colon there. I'm laughing right now. 🤣 💛 
Reviews please. 😃

lib/get-request-agent.js
```diff
@@ -20,28 +20,29 @@ function getRequestAgent (options) {
   if (agentOptionNames.length === 0) {
     return
   }
-
```
gr2m, line 23: nitpick: avoid changes unrelated to the PR :)

test/integration/params-validations-test.js
```diff
@@ -22,8 +22,26 @@ describe('params validations', () => {
...
+  })
+
+  it('coverage', () => {
+    const github = new GitHub({
+      baseUrl: null,
       host: '127.0.0.1',
-      port: 8 // officially unassigned port. See https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers
+      port: '8',
+      protocol: 'https'// port: 8 // officially unassigned port. See https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers
     })
```
gr2m: remove the coverage test here, we should cover it all in `deprecations-test.js`

test/integration/smoke-test.js
```diff
@@ -15,9 +15,22 @@ describe('smoke', () => {
...
+      .get('/my/api/orgs/myorg')
+      .reply(200, {})
+
+    const github = new GitHub({
       protocol: 'http',
-      pathPrefix: '/my/api/'
+      host: 'myhost.com',
+      pathPrefix: '/my/api',
+      baseUrl: null
     })
```
gr2m: remove the `protocol, port, pathPrefix coverage` test here, we should cover it all in `deprecations-test.js`

lib/parse-client-options.js
```diff
@@ -7,6 +7,7 @@ const getRequestAgent = require('./get-request-agent')
 const DEFAULTS = require('./defaults')
 const OPTION_NAMES = [
   'timeout',
+  'baseUrl',
```
gr2m: remove 

```js
   'host',
   'pathPrefix',
   'protocol',
   'port',
```

instead we should check for them "manually" below. That way it will be easier to remove them in future, as they are implemented in only one block

lib/parse-client-options.js
```diff
@@ -38,18 +55,22 @@ function parseOptions (userOptions) {
...
-  if (options.pathPrefix) {
-    clientDefaults.baseUrl += '/' + options.pathPrefix.replace(/(^[/]+|[/]+$)/g, '')
+    // Check if a prefix is passed in the options and strip any leading or trailing slashes from it.
+    if (options.pathPrefix) {
+      clientDefaults.baseUrl += '/' + options.pathPrefix.replace(/(^[/]+|[/]+$)/g, '')
+    }
   }
-
   /* istanbul ignore else */
+
```
gr2m, line 67: let’s implement this a little different. First, let’s set `clientDefaults.baseUrl` directly in line 51

```js
  const clientDefaults = {
    baseUrl: options.baseUrl,
    headers: options.headers,
    request: {
      timeout: options.timeout
    }
  }
```

Because that’s the default path now, and once we remove the deprecated options, we can just leave it as is.

Next instead of

```js
  if (options.baseUrl) {
    clientDefaults.baseUrl = `${options.baseUrl}`
  } else {
```

I would suggest to check for `if (userOptions.host) { ...` because the host options must always be set. Then within the if block, replace `options.*` with `userOptions.*`, because we directly want to check for what the user passed, without the default values applied to them and without being filtered by `OPTION_NAMES` defined on top.

test/integration/deprecations-test.js
```diff
@@ -33,6 +33,34 @@ describe('deprecations', () => {
     expect(console.warn.callCount).to.equal(1)
   })
 
+  it('deprecated protocol option', () => {
+    GitHub({
+      protocol: 'https'
+    })
+    expect(console.warn.callCount).to.equal(1)
+  })
```
gr2m: add `host: 'deprecations-test.com'` option to all 4 tests, because just setting `protocol` is no valid use case. That will also fix the coverage

gr2m (commented):
great work @aryannarora 👍 Now I just have some stylistic comments, I hope you don’t mind. The PR is great as is and I’m happy to make the final clean ups myself if you are too busy, just let me know :)

test/integration/params-validations-test.js
aryannarora: Should i move this entire test to deprecations-test.js ?

lib/get-request-agent.js
aryannarora: Oops. I'll remember it. 😃

lib/parse-client-options.js
aryannarora: Sure.

test/integration/params-validations-test.js
gr2m: you can remove it entirely, we don’t need it, you’ll just need to adjust the deprecations test and the implementation a bit, see my other comments

aryannarora:
@gr2m Everything's done. 
Review please. 🙂

lib/parse-client-options.js
```diff
@@ -54,7 +54,7 @@ function parseOptions (userOptions) {
 
   if (userOptions.baseUrl) {
     clientDefaults.baseUrl = `${userOptions.baseUrl}`
-  } else {
+  } else if (userOptions.protocol) {
```
gr2m: you can replace

```js
  if (userOptions.baseUrl) {
    clientDefaults.baseUrl = `${userOptions.baseUrl}`
  } else if (userOptions.protocol) {
```

with just 

```js
if (userOptions.protocol) {
```

aryannarora:
@gr2m I really should've noticed that. 🙈 
Did it now. :)

gr2m (approved):
fabulous 👍

aryannarora:
@gr2m So great. ❤️

PR #814, opened 2018-03-13
fix: document default value for issues.getAll endpoint

### LINKED ISSUE #639: issues.getAll should document the default for filter
The default should be documented, i assumed it was "all", but it looks like it is something else.

screen shot 2017-10-28 at 1 52 25 pm

synicalsyntax:
Fixes #639

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -2133,7 +2133,9 @@
             "created",
             "mentioned",
             "subscribed"
-          ]
+          ],
+          "default": "assigned",
+          "description": "Which sort of issue to return."
         },
         "state": {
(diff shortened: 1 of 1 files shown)


gr2m (approved):

gr2m:
Thanks a lot @synicalsyntax! Just for your information: as the code does not change any code but only documentation, I usually use the `docs: ...` prefix. With a `fix: ...` commit, a new version would get released automatically which is not necessary. 

I can change the commit easy enough with squash & merge, just wanted to let you know about it as this is not your first pull request. Keep ’em coming, they are 💯

PR #815, opened 2018-03-15
Update types to reflect baseUrl changes

tcbyrd:
@gr2m @philschatz I think this is the change you need to the type definitions to reflect the deprecations in https://github.com/octokit/rest.js/pull/807

See https://github.com/octokit/rest.js/pull/732#issuecomment-373554225 for more context on where I ran into this.

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -38,14 +38,38 @@ declare namespace Github {
     requestMedia?: string;
     agent?: http.Agent;
 
-    /** deprecated */
+    /**
+     * @deprecated in version 15.0.0
+     */
     proxy?: string;
+    /**
+     * @deprecated in version 15.0.0
+     */
     ca?: string;
+    /**
+     * @deprecated in version 15.0.0
+     */
     rejectUnauthorized?: boolean;
+    /**
+     * @deprecated in version 15.0.0
+     */
     family?: number;
+
+    /**
+     * @deprecated in version 15.2.0
+     */
     host?: string;
+    /**
+     * @deprecated in version 15.2.0
+     */
     pathPrefix?: string;
+    /**
+     * @deprecated in version 15.2.0
+     */
     protocol?: string;
+    /**
+     * @deprecated in version 15.2.0
+     */
     port?: number;
   }
 
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -32,11 +32,8 @@ declare namespace Github {
   }
 
   export interface Options {
+    baseUrl?: string;
     timeout?: number;
-    host?: string;
-    pathPrefix?: string;
-    protocol?: string;
-    port?: number;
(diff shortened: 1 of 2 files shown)


philschatz:
@tcbyrd Yes that should fix it. #803 added a `test/typescript-validate.ts` to help catch changes that would require a change in the TypeScript definition.

It might be helpful to include all of the optional arguments to the constructor so that they are updated in the TypeScript definition when they are updated in the JavaScript.

scripts/templates/index.d.ts.tpl
```diff
@@ -32,11 +32,8 @@ declare namespace Github {
   }
 
   export interface Options {
+    baseUrl?: string;
     timeout?: number;
-    host?: string;
-    pathPrefix?: string;
-    protocol?: string;
-    port?: number;
```
gr2m, line 39: I think I would leave these, they are deprecated but still supported. That way we don’t break existing code.

Is there a way to mark something as "deprecated" in Typescript?

tcbyrd: I think one way is to [decorate](http://usejsdoc.org/tags-deprecated.html) it with `@deprecated` so that you get the message from the language server. Not sure if this causes a warning in the build.

gr2m: Mind if I update your PR? I’ve extended the typescript test to make it less likely that I break it again :(

tcbyrd: Be my guest!

tcbyrd: JSDoc doesn't cause a build failure, but it does do this:

image

when you do this:
```
/**
  * @deprecated in version 15.2.0
  */
host?: string;
```

gr2m: I don’t use JS doc in this repo, is this an editor plugin?

tcbyrd: >  is this an editor plugin?

It's part of the VS Code / Atom language server for TypeScript

gr2m: okay thanks, I added them to each deprecated option

gr2m:
Does the PR look good to you now @philschatz @tcbyrd 

Btw if you have an app depending on `@octokit/rest` that is using Typescript that you’d like me to test with before doing releases, I’d be happy to add it to my local test setup that I usually test against when working on fixes / features before releasing

tcbyrd:
@gr2m I think once if we finally get Probot switched over, that could be a great way to test. The compiler should help catch things like this.

tcbyrd:
> Does the PR look good to you now @philschatz @tcbyrd

Unless @philschatz has any other suggestions on deprecations, I'd say it's good to 🚢

philschatz:
Yep, the JSDoc comment is the only way I know of to add deprecation notices to TypeScript files. 👍 for merging.

gr2m (approved):

Issue #833, opened 2018-04-06
activity.getNotifications date parsed into incorrect format


JuanCaicedo:
I'm calling `octokit.getNotifications({ since: '2018-04-06T20:04:47.000Z'})` and am finding that it makes a request to 
> <https://api.github.com/notifications?since=Fri%20Apr%2006%202018%2015%3A04%3A47%20GMT-0500%20(-05)>
 
As far as I can tell, this is incorrect, the request should be made to 
> <https://api.github.com/notifications?since=2018-04-06T20:04:47.000Z>

I say that because in the [api docs](https://developer.github.com/v3/activity/notifications/#list-your-notifications) it says `since` should be an ISO string.

I think the root cause is the same as https://github.com/octokit/rest.js/issues/716, but a regression was introduced at some point. My reasoning is if you add this code 
```js
  console.log('options', options.since instanceof Date)
``` 
on line 15 at https://github.com/octokit/rest.js/blob/master/lib/plugins/endpoint-methods/method.js#L15, it will evaluate to true.

I'm still tracking down the root cause of that problem. I think `since` should be a string, but it remains a date even if I change `routes.json`

gr2m:
Thanks for looking into it @JuanCaicedo. If you could start a pull request that reproduces the error, that’d be a great start.

JuanCaicedo:
@gr2m Do you mean to add a failing test based on this error? If so sure 🙂

JuanCaicedo:
Looking at how other date parameters work, like `due_on` for `issues.createMilestone()`, it looks like they should be an instance of `Date`, not a string.

I suspect that the difference is that in the `due_on` case, that date value is stringified and put into the body, while in the `since` case, it is url encoded and put into the url

JuanCaicedo:
Sorry, I think this is my error in how I was calling the library. I was passing a date object, and what you should pass is an ISO string like 
```js
const date = new Date('2018-04-06T20:04:47.000Z')
octokit.getNotifications({ 
  since: date.toISOString()
})
```
I think the real fix for the problem I experienced would be to validate that the date parameter is a valid ISO string. In https://github.com/octokit/rest.js/pull/719, I just removed the validation completely. Not sure of how to validate an ISO string, but it might be as easy as just trying to make a new date object with it and throwing an error if that fails.

Let me know if you think that's a good path forward and I can open up another issue about it and work on implementing it

gr2m:
[...]

PR #834, opened 2018-04-09
failing test for notifications date parsing

JuanCaicedo:
Adds a failing test to showcase the problem in #833 

Note that this also causes a second test to fail that expects there not to be any pending mocks. I think this should go away once the test is passing

### DIFF
--- a/test/integration/params-validations-test.js
+++ b/test/integration/params-validations-test.js
@@ -127,4 +127,17 @@ describe('params validations', () => {
       due_on: new Date('2012-10-09T23:39:01Z')
     })
   })
+
+  it('Date object for github.activity.getNotifications({..., since})', () => {
+    const github = new GitHub({
+      host: 'notifications-test-host.com'
+    })
+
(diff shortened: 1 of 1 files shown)


test/integration/params-validations-test.js
```diff
@@ -127,4 +127,17 @@ describe('params validations', () => {
...
+    const github = new GitHub({
+      host: 'notifications-test-host.com'
+    })
+
+    nock('https://notifications-test-host.com')
+      .get('/notifications')
+      .query({ since: '2012-10-09T23:39:01Z' })
+      .reply(201, {})
+
+    return github.activity.getNotifications({since: new Date('2012-10-09T23:39:01Z'})
```
JuanCaicedo: I'm a little unclear on if `since` should be a date or a string. Based on the fact that `due_on` is a date object in the above example, I figured it should be here as well

JuanCaicedo:
Sorry, I need to resolve some problems here first 😅

PR #842, opened 2018-04-28
handle empty responses to HEAD requests

### LINKED ISSUE #841: Testing if a pull request exists
I'd like to test if a pull request exists. For example, a function that returns true for this pull request: https://github.com/whatwg/html/pull/3645 and false for this pull request: https://github.com/whatwg/html/pull/3594.

Ideally I would like to do this by inputting a URL, instead of parsing the URL to pull out owner/repo/number. (I am checking whether a pull request pointed to via someOtherResponse.pull_request.url exists.) But that's a side note.

I couldn't figure out a good way to do this within the confines of this package. Things I tried:

- Using pullRequests.get
  - Will fetch the entire response; quite wasteful
- Using pullRequests.get with method: "HEAD"
  - Gives an unhandled promise rejection warning about invalid JSON body, presumably since an empty body is not valid JSON
- Using the internal request module: https://github.com/octokit/rest.js/tree/master/lib/request
  - I don't want to depend on unstable functionality
- Dropping down to node-fetch manually
  - Now I have to assemble my auth tokens etc. myself, losing much of the benefits of this library.

Any ideas?

gr2m:
fixes #841. Does that resolve your problem @domenic?

### DIFF
--- a/lib/request/request.js
+++ b/lib/request/request.js
@@ -39,14 +39,21 @@ function request (requestOptions) {
   return fetch(requestOptions.url, pick(requestOptions, 'method', 'body', 'headers', 'timeout', 'agent'))
 
     .then(response => {
-      const contentType = response.headers.get('content-type')
-
       for (const keyAndValue of response.headers.entries()) {
         headers[keyAndValue[0]] = keyAndValue[1]
       }
 
       if (response.status === 204) {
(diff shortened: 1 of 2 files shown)


domenic:
This looks great, thank you!

Issue #727, opened 2018-01-30
Unexpected end of JSON input


iamnewton:
I'm pretty sure I'm doing something wrong here, but I've exhausted what it is.

```
const octokit = require('@octokit/rest')({
    debug: true,
    host: 'github.<company domain>.com',
    pathPrefix: '/api/v3'
});

octokit.authenticate({
    type: 'token',
    token: '<personal access token with all scopes'
});

octokit.orgs.get({org: '<valid org>'}, (error, result) => {
    console.log(error);
});
```

Output:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at IncomingMessage.response.on (./node_modules/@octokit/rest/lib/request/request.js:79:31)
    at IncomingMessage.emit (events.js:164:20)
    at endReadableNT (_stream_readable.js:1054:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)```

**Comment by gr2m** [comment]
the `debug` option has been removed in an earlier release. Could you run the script with `DEBUG=octokit*` and paste the logs here?

Just to make sure I tried to reproduce the problem against GitHub’s API (default setting) and with that it all looks good: https://runkit.com/gr2m/5a710542399a8600123e8f5f

**Comment by iamnewton** [comment]
```
#19445: $ node bin/set-status.js
  octokit:rest REQUEST: Url {
  protocol: 'https:',
  slashes: true,
  auth: null,
  host: 'github.<company domain>.com',
  port: null,
  hostname: 'github.<company domain>.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/api/v3/orgs/<valid org>',
  path: '/api/v3/orgs/<valid org>',
  href: 'https://github.<company domain>.com/api/v3/orgs/<valid org>',
  method: 'get',
  body: undefined,
  headers:
   { accept: 'application/vnd.github.v3+json',
     'user-agent': 'octokit/rest.js v14.0.7',
     authorization: 'token <personal access token with all scopes>' },
  ca: undefined,
  family: undefined,
  proxy: undefined,
  rejectUnauthorized: true,
  timeout: 0 } +0ms
  octokit:rest STATUS: 200 +264ms
[...]

PR #853, opened 2018-05-08
Enterprise & baseUrl option

gr2m:
test for #727

### DIFF
--- a/test/issues/727-enterprise-base-url.js
+++ b/test/issues/727-enterprise-base-url.js
@@ -1,0 +1,22 @@
+const nock = require('nock')
+const Octokit = require('../../')
+
+require('../mocha-node-setup')
+
+describe('https://github.com/octokit/rest.js/issues/727', () => {
+  it('sends request to correct baseUrl', () => {
+    nock('https://github.domain.com')
+      .get('/orgs/frontendgine/repos')
+      .query({type: 'public'})
+      .reply(200, [])
+
+    const client = new Octokit({
+      baseUrl: 'https://github.domain.com'
+    })
+
+    client.repos.getForOrg({
+      org: 'frontendgine',
+      type: 'public'
+    })
+  })
+})


gr2m:
@chrisopedia I tried to reproduce the problem but no luck, it sends the request correctly to the URL set with the `baseUrl` option

iamnewton:
Is there something with my code and/or setup then?

gr2m:
I really don’t know, it’s puzzling. 

Are you still on GHE 2.11? Are you using the latest `@octokit/rest`?

I just tested it with GHE 2.12 (the only version I have available for testing right now) and the latest `@octokit/rest`, no error.

Do you maybe use a proxy in your stack? Do you set the `agent` option? Do you set any before/after hooks in your code?

iamnewton:
> Are you still on GHE 2.11? Are you using the latest @octokit/rest?

yes, still on GHE 2.11. And fresh install of `rest`

> Do you maybe use a proxy in your stack? 

no, I have to go through a VPN

> Do you set the agent option? 

no

> Do you set any before/after hooks in your code?

no

gr2m:
I now tested it against a freshly setup GHE v2.11.16 (revision ab9a643), I could not reproduce the problem.

Let’s try to get some better debug information.

Can you change this line directly in your setup:

https://github.com/octokit/rest.js/blob/65c3a2942caf991d55ba6c89449f54f5e9ee53ef/lib/request/request.js#L74

I’m pretty sure that this is were the issue occurs.

Can you change it to

```js
return response.text()

  .then(text => {
    console.log('SERVER RESPONSE:')
    console.log(text)
    return JSON.parse(text)
  })
```

gr2m:
Is this all of the response you get?

```
core.com:frontendgine/onyx.git","clone_url":"https://github.rp-core.com/frontendgine/onyx.git","svn_url":"https://github.rp-core.com/frontendgine/onyx","homepage":null,"size":0,
```

Do you have a firewall or anything else between the GHE server and the place you send the request from?

iamnewton:
I get a lot more from the response, just didn't want to paste it all in.  Looks like the `JSON` parsing is whats screwing it up.  I'm not sure we do, because I can `curl` the API just fine.

gr2m:
can you paste it somewhere safe and send me the link, or the entire response, to octokit@martynus.net?

The JSON parsing is part of https://github.com/bitinn/node-fetch which is pretty well tested and widely used.

Maybe try to send the request with node-fetch directly and see if that works?

iamnewton:
Sent to your email.

gr2m:
no luck with reproducing on my side either.

Can you try this

```js
const fetch = require('node-fetch')
const url = 'https://github.rp-core.com/api/v3/orgs/frontendgine/repos?type=public'
const token = '...'
fetch(url, {
  method: 'get',
  headers: {
    accept: 'application/vnd.github.v3+json',
    'user-agent': 'octokit/rest.js v0.0.0-semantically-released',
    authorization: `token ${token}`
  }
})

  .then(response => response.json())

  .then(data => {
    console.log(`\ndata.length ==============================`)
    console.log(data.length)
  })
```

iamnewton:
```
(node:98568) UnhandledPromiseRejectionWarning: FetchError: invalid json response body at https://github.rp-core.com/api/v3/orgs/frontendgine/repos?type=public reason: Unexpected end of JSON input
    at /Users/chrisopedia/Code/frontendgine/legion/develop-cli/node_modules/node-fetch/lib/index.js:254:32
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:118:7)
(node:98568) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:98568) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

gr2m:
You said you could `curl` the API just fine. The response that you get, is that valid JSON, or is it also cut off?

I very much looks like it’s not a client problem, but a server problem. The response is cut off where it should not

iamnewton:
yeah, I just sent you via email the `curl` response.

gr2m:
Could you send me the full curl command with the `-i` flag, including the response with headers? Just want to make sure you set the `accept`. I’ve seen cut off responses before, but it was always a server problem where the server calculated the content length incorrectly. But I don’t think that GHE enterprise would do that, it would need to be sth else in between.

I’m out of ideas :( I’d suggest to take it to node-fetch, maybe they can help you there. It might be a problem with `node-fetch`. I’d be happy to open the issue there for you, I’m helping maintain it.

iamnewton:
Sent.  Can you send me an example of the command you're using that is successful?  I'd like to try it without it going through my GHE to ensure that its not my machine.

iamnewton:
yeah, if you could open up an issue on node-fetch that would be awesome.

gr2m:
closing, as discussed with @chrisopedia.

PR #855, opened 2018-05-08
update routes

gr2m:
closes #854


gr2m:
I gotta figure out the right syntax to use environment variables in script commands :/ 
- https://travis-ci.org/octokit/rest.js/jobs/376439881#L4031
- https://github.com/octokit/rest.js/blob/65c3a2942caf991d55ba6c89449f54f5e9ee53ef/.travis.yml#L66

Issue #830, opened 2018-04-01
webpacked to Lambda: fetch undefined


wyvern8:
Hi @gr2m, hope you are well.   Looks like we are seeing issues with the new node-fetch usage in request.js on Lambda after webpack via serverless framework.  The 'fetch' function is not defined it seems.  As per the previous issue transpiled to Lambda, a targeted reproduction will take a little time,  but letting you know in case others report.  

I tried a few angles in webpack conf based on similar isssues, but have not found a suitable solution as yet, so rolled back to 15.1.3 for now.  I'm wondering if it is to do with webpack not bundling transitive dependencies correctly due to the runtime bindings that happen as part of https://github.com/gr2m/before-after-hook

gr2m:
I think this might have been a problem with node-fetch https://github.com/bitinn/node-fetch/commit/d522036bee580abdca0fc6ab7afcd38f573e4f96

I should be fixed in the latest version:  2.1.2. Can you make sure you have that one? You can check with `npm ls node-fetch`

Cethy:
Bite me too.
BTW, I tried using [philschatz/octokat.js](https://github.com/philschatz/octokat.js/) to found the same (kinda) problem. So it should come from node-fetch (or webpack).

I have `node-fetch@2.1.2` (for octokit) and it do not fix the problem.
Reverting to `octokit@15.1.3` do remove the dependency to `node-fetch` and works.

prabhu:
Are you back from vacation mate @gr2m ?

a-marquez:
This seems to affect parcel too. I'm reverting as well for now, hopefully it'll be the same fix.

gr2m:
How would you test for that? I’d be good to have a test as part of our CI so we don’t break it again in future. And it would help me to reproduce the problem without setting up lambda, which I’m not using right now

gr2m:
I’ve setup a lambda function and deployed using the `aws-cli` as described at https://stackoverflow.com/questions/34437900/how-to-load-modules-in-aws-lambda

It worked without a problem.

@wyvern8 @prabhu Can you share a minimal webpack setup that you use when you run into a problem?

### Function

```js
const Octokit = require('@octokit/rest')

exports.handler = async (event) => {
    const octokit = Octokit()
    return octokit.request({
        method: 'GET',
        url: '/'
    })
};
```

### package

```json
{
  "name": "octokit-rest-lambda",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
[...]

PR #858, opened 2018-05-10
refactor: replace 'node-fetch' dependency with 'cross-fetch' 

a-marquez:
This patch replaces the `node-fetch` dependency with `cross-fetch`. Please note both `node-fetch` and `@gr2m/node-fetch` are removed from package.json.

This may address the "fetch is undefined" error from #830.

### DIFF
--- a/lib/request/request.js
+++ b/lib/request/request.js
@@ -2,7 +2,7 @@
 
 module.exports = request
 
-const fetch = require('node-fetch')
+const fetch = require('cross-fetch')
 const debug = require('debug')('octokit:rest')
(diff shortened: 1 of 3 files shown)


gr2m:
The bundle size increases quite a bit, from 23.16KB to 24.95KB.

To check, run this locally

```
npm run build:browser
npx bundlesize
```

It might not look like much right now because we are above 20kb, but I think we can get it under 10kb and then it will be quite significant

gr2m:
Also looks like the browser tests are failing?

Issue #860, opened 2018-05-11
Checks annotations and check images do not work.


monder:
When trying to create a new check with annotations or images error is thrown:
```
HttpError: Empty value for parameter 'output.annotations.blob_href': undefined
```

Seems the definition of routes is wrong [here](https://github.com/octokit/rest.js/blob/b716c30ff1ed78b16813cb094bcc2c80445cee05/lib/routes.json#L1043) it should be `output.annotations[].blob_href` instead of `output.annotations.blob_href`.
The same for `output.images` [here](https://github.com/octokit/rest.js/blob/b716c30ff1ed78b16813cb094bcc2c80445cee05/lib/routes.json#L1078)

gr2m:
Thanks @monder I’m looking into it

gr2m:
sorry did not yet land in `@octokit/rest`, on its way

PR #864, opened 2018-05-11
Custom Accept header for checks APIs

### LINKED ISSUE #861: Custom Accept header not provided to Checks API
Using `@octokit/rest` version `15.4.0`, using [`checks.create`](https://github.com/octokit/rest.js/blob/b716c30ff1ed78b16813cb094bcc2c80445cee05/lib/routes.json#L997), the [custom `Accept` header](https://developer.github.com/changes/#how-can-i-try-it) (`application/vnd.github.antiope-preview+json`) isn't provided, resulting in an error in the following code:

```javascript
octokit.checks.create({
  owner: 'swinton',
  repo: 'example',
  name: 'feedback',
  head_branch: 'helloworld',
  head_sha: '2e3d00a6f14a667d50ad9ccd6f3dcfded52121e2',
  status: 'in_progress',
  started_at: (new Date()).toISOString(),  // The time that the check run began in ISO 8601 format: `YYYY-MM-DDTHH:MM:SSZ`
  completed_at: (new Date()).toISOString(),
  conclusion: 'status'
})
```

The error returned is:

> If you would like to help us test the Checks API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details.

It would be nice if the accept header was included by default :bow:

### LINKED ISSUE #865: When authenticated as integration, accept header gets always set to application/vnd.github.machine-man-preview+json
follow up for #864 

https://github.com/octokit/rest.js/blob/72bd9afe7c8eb0a82fef3f229d5ab6c75d04f112/lib/plugins/authentication/before-request.js#L21-L25

It should amend the accept header, multiple accept headers can be passed, separated by commas

gr2m:
fixes #861 

### TODOs

- fix #865, then remove `application/vnd.github.machine-man-preview+json` from default headers

### DIFF
--- a/lib/plugins/authentication/authenticate.js
+++ b/lib/plugins/authentication/authenticate.js
@@ -21,6 +21,7 @@ function authenticate (state, options) {
 
     case 'token':
     case 'integration':
+    case 'app':
       if (!options.token) {
         throw new Error('Token authentication requires a token to be set')
       }
--- a/lib/plugins/authentication/before-request.js
+++ b/lib/plugins/authentication/before-request.js
@@ -18,7 +18,13 @@ function authenticationBeforeRequest (state, options) {
     return
   }
 
+  // deprecate state.auth.type === 'integration', rename to 'app'
   if (state.auth.type === 'integration') {
+    console.warn('authentication type "integration" is deprecated. Use "app" instead.')
+    state.auth.type = 'app'
+  }
+
+  if (state.auth.type === 'app') {
     options.headers['authorization'] = `Bearer ${state.auth.token}`
     options.headers['accept'] = 'application/vnd.github.machine-man-preview+json'
     return
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -995,6 +995,9 @@
   },
   "checks": {
     "create": {
+      "headers": {
+        "accept": "application/vnd.github.antiope-preview+json,application/vnd.github.machine-man-preview+json"
+      },
       "method": "POST",
       "params": {
         "completed_at": {
@@ -1123,6 +1126,9 @@
       "url": "/repos/:owner/:repo/check-runs"
     },
     "get": {
+      "headers": {
+        "accept": "application/vnd.github.antiope-preview+json,application/vnd.github.machine-man-preview+json"
+      },
       "method": "GET",
       "params": {
         "check_run_id": {
@@ -1141,6 +1147,9 @@
       "url": "/repos/:owner/:repo/check-runs/:check_run_id"
     },
     "getSuite": {
+      "headers": {
+        "accept": "application/vnd.github.antiope-preview+json,application/vnd.github.machine-man-preview+json"
+      },
       "method": "GET",
       "params": {
         "check_suite_id": {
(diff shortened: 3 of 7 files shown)


gr2m:
@swinton I created a test to make sure that you can set custom accept headers with both `{headers: {Accept: '...'}}` and `{headers: {accept: '...'}}` and couldn’t find a problem there. You mentioned that this didn’t work for you?

I’m now adding the default preview header to the new checks routes so this becomes a non-issue, but still wanted to check on the custom header setting because that needs to work for other use cases

gr2m:
@swinton @monder can you check if this PR fixes the problem for you?

monder:
@gr2m I've checked my issue and everything works perfectly, but I'm always setting the header. Issue 
https://github.com/octokit/rest.js/issues/860 is fixed in 15.4.1. Thanks for the quick response.
@swinton Indeed, its a bit confusing and took me a couple of hours to realize that these tokens are not the same. You can only use `integration` with [Apps api](https://octokit.github.io/rest.js/#api-Apps).
Basically you have to do something like this:
```javascript
const githubToken = jwt.sign({
    iat: Math.floor(Date.now() / 1000),
    exp: Math.floor(Date.now() / 1000) + 9 * 60,
    iss: `${config.github.iss}`,
}, config.github.key, { algorithm: 'RS256' });

octokit.authenticate({
    type: 'integration',
    token: githubToken,
});

const tokenResult = await octokit.apps.createInstallationToken({installation_id: config.github.installation_id})
octokit.authenticate({
    type: 'token',
    token: tokenResult.data.token,
});

const body = {
    owner: 'monder',
    repo: 'experiments',
    head_branch: 'master',
    head_sha: '308ccfb77bb12024e1ea69f61374b3051f13cf71',
    name: 'test_check',
    status: 'completed',
    completed_at: new Date().toISOString(),
    conclusion: 'failure',
    output: {
        title: 'Failure',
        summary: '',
        annotations: [{
            filename: 'README.md',
            blob_href: `https://github.com/monder/experiments/blob/308ccfb77bb12024e1ea69f61374b3051f13cf71/README.md`,
            warning_level: 'failure',
            message: 'message text',
            start_line: 2,
            end_line: 2,
        }],
    },
};

const res = await octokit.checks.create({
    ...body,
    headers: {
        accept: 'application/vnd.github.antiope-preview+json'
    }
});
```
Note that the differ `installation_id` from `iss` and in my case I didn't want to waste api calls so I just hardcoded it for our organisation.
And also FYI, annotations do not show up on github when viewing PR in `split` mode, only `unified` view works for now.

swinton:
> you can set custom accept headers with both {headers: {Accept: '...'}} and {headers: {accept: '...'}} and couldn’t find a problem there. You mentioned that this didn’t work for you?

This is working for me now, thanks to [the guidance](https://github.com/octokit/rest.js/pull/864#issuecomment-388488189) from @monder 

Basically, in [my code above](https://github.com/octokit/rest.js/pull/864#issuecomment-388469653), I was passing the wrong authentication type (`integration` instead of `token` 🤦‍♂️) I changed that and it works :tada:

PR #870, opened 2018-05-14
amend accept header for github apps, keep custom accept headers intact

gr2m:
closes #864


gr2m:
I’ve run into failing tests with Greenkeeper’s test suite, investigating

bkeepers:
This is causing an issue for any requests that are [authenticated as an installation](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation) and accessing these endpoints that require the `machine-man-preview` header.  Installation tokens use the same format as other tokens: `Authorization: token <installation-token>`.

PR #873, opened 2018-05-15
Add the apps bit to the types

orta:
Adds type definitions to the changes in https://github.com/octokit/rest.js/pull/866

There's not really a way to deprecate something in TypeScript/flow so I just added a note in the TS one, because it'll show up in autocomplete.


gr2m (approved):
Thanks 👍

PR #875, opened 2018-05-15
Fix HTML typo

DavidBruant:


gr2m (approved):
Great catch, thanks a lot!

PR #878, opened 2018-05-19
machine-man-preview header required for installation tokens

bkeepers:
https://github.com/octokit/rest.js/pull/870 introduced a regression where a client authenticated with an app installation token doesn't get the `machine-man-preview`.

This adds a test to expose the issue, and reverts the route changes.

cc https://github.com/probot/probot/issues/537

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -523,6 +523,9 @@
   },
   "apps": {
     "addRepoToInstallation": {
+      "headers": {
+        "accept": "application/vnd.github.machine-man-preview+json"
+      },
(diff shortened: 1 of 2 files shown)


gr2m:
looking into it

orta:
FWIW, I think I'm hitting this with Peril too, just adding some debugging tools to verify

gr2m:
I’m gonna go ahead and merge it as it fixes the problem Brandon reported and has a very low risk of breaking new things. @orta if you could give it a try that’d be great.

PR #892, opened 2018-05-26
Update @octokit/routes to version 8.0.0

gr2m:


igrayson:
Was a breaking change? It appears to add several new required parameters to `routes.json`.

gr2m:
The parameters where required before, GitHub’s would have responded with an error rather then the client before sending the request. So it was a fix more than a breaking change

igrayson:
The following typescript code snippet works in `15.4.1` (no error from github), but does not compile with `15.8.0` (missing required `comment_id`).
```ts
  await octokit.issues.editComment({
    owner: repo.owner,
    repo: repo.name,
    id: commentId.toString(),
    body: newText,
  });
```

Workaround code to support `15.8.0`:
```ts
  await octokit.issues.editComment({
    owner: repo.owner,
    repo: repo.name,
    // argument duplicated; ref https://github.com/octokit/rest.js/pull/892
    id: commentId.toString(),
    comment_id: commentId.toString(),
    body: newText,
  });
```

gr2m:
Typescript is not a production dependency so I wouldn’t consider it a breaking change. It’s not breaking the compiled JS code.

Can you just set “comment_id”?

PR #893, opened 2018-05-26
log deprecation messages only once

### LINKED ISSUE #871: Log deprecation messages only once and add "@octokit/rest" to the message
No need to log a deprecation message twice. That will avoid excessive logging that looks like this

image

gr2m:
closes #871


gr2m:
Problem is that depd would only log once depending of the location of the log statement, not the message. So if people would use 10 deprecated parameter names they would only see one message

PR #894, opened 2018-05-27
s/namspaced/namespaced

shiftkey:
Noticed this typo while wrapping my head around how types are generated from the routes source.


gr2m (approved):
Thanks for being a good Open Source Citizen 👍

PR #895, opened 2018-05-27
Don’t log same deprecation message twice

### LINKED ISSUE #871: Log deprecation messages only once and add "@octokit/rest" to the message

gr2m:
closes #871. Follow up for https://github.com/octokit/rest.js/pull/893#issuecomment-392303982

### DIFF
--- a/lib/deprecate.js
+++ b/lib/deprecate.js
@@ -1,0 +1,12 @@
+module.exports = deprecate
+
+const loggedMessages = {}
+
+function deprecate (message) {
+  if (loggedMessages[message]) {
+    return
+  }
+
+  console.warn(`DEPRECATED (@octokit/rest): ${message}`)
+  loggedMessages[message] = 1
+}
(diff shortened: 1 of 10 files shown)


lib/deprecate.js
```diff
@@ -0,0 +1,12 @@
...
+const loggedMessages = {}
+
+function deprecate (message) {
+  if (loggedMessages[message]) {
+    return
+  }
+
+  console.warn(`DEPRECATED (@octokit/rest): ${message}`)
+  loggedMessages[message] = 1
+}
```
gr2m, line 12: this stores logged messages globally in memory. If `@octokit/rest` is instantiated multiple times then a message logged by a previous instance will not be logged again, which I think is right behaviour here. And there is not a high risk of a memory leak because there is a finite number of unique deprecation messages

PR #910, opened 2018-06-10
Async Iterators for paginated results

### LINKED ISSUE #688: Autopaginate
Is there a way to autopaginate requests?, this would be useful in some cases
  
for example you want to get all repositories of an organization and the organization has more than 100 repositories

gr2m:
closes #688

### Update Nov 7

This is the API I think we should implement:

```js
const options = github.issues.getForRepo.endpoint.merge({owner, repo})

// returns array of all issues across all pages
const issues = await github.paginate(options)

// map results
const issueTitles = await github.paginate(options, ({ data }) => data.issues.title)

// async iterator
for await (const response of github.paginate.iterator(options) {
  // do whatever you want with each response, break out of loop, etc.
}
```

See https://github.com/octokit/rest.js/pull/910#issuecomment-436871797

---

<details>
  <summary>Outdated</summary>
For Node < 10 and browsers without support for async iterators, the code is not overly elegant:

```js
const iterator = github.issues.getForRepo(options)[Symbol.asyncIterator]()
const results = []
let result

while (true) {
  result = await iterator.next()
  if (result.done) break
  results.push(...result.value.data)
}
```

We will probably create an officially supported pagination plugin to hide away the complexity. The advantage is that calling the `.next()` method is easier than keeping state of the last response’s link header, we can contain that logic and improve it in future without API changes.

For Node 10 and modern browsers, paginating is pretty neat

```js
const results = []
for await (const result of github.issues.getForRepo(options)) {
  results.push(...result.data)
}
```

It’s implemented by decorating the returned promise with a `[Symbol.asyncIterator]` property

https://github.com/octokit/rest.js/blob/33fa14d8498678189eba49090a4ce7ae56749fbb/lib/plugins/endpoint-methods/method.js#L33-L44
</details>


test/scenarios/paginate-issues-test.js
```diff
@@ -0,0 +1,86 @@
...
+
+        return iterator.next()
+      })
+      .then(result => {
+        expect(result.done).to.equal(true)
+      })
+  })
+
+  // This test will only run in Node 10 and browsers that support async iterators
+  // it('for await (let result of github.issues.getForRepo()', async () => {
```
bkeepers: Do you have ideas how to make this run conditionally?

bkeepers (commented):
This is some mad science, but I think I like it.

I think it would be worth documenting how to do this in the README or somewhere, along with examples for a few different use cases:

- paginate through all results (the current example)
- paginate until I find the thing I'm looking for
- paginate and map a field from the results into a single resulting array

test/scenarios/paginate-issues-test.js
gr2m: I tried, but couldn’t find a simple way yet. I’m sure there is, I just didn’t want to spend more time on it right now

gr2m:
There was some good feedback about the API on twitter: https://twitter.com/gr2m/status/1005636713933258752

This API would not allow to create a dozen of iterators, each call would send a request before the first iteration, see e.g. https://twitter.com/reconbot/status/1006251185915998209. But I’m not sure if that’s a problem in this case.

The API looks really neat to me but I got bit by this kind of "magic" in the past which is why I’m reluctant. To avoid side effects we could do something like this

```js
for await (const result of client.issues.getAll.pages()) {
  // do whatever you like
}
```

Instead of 

```js
for await (const result of client.issues.getAll()) {
  // do whatever you like
}
```

Less magic, but dunno if that would be easier to explain 🤔 

> I think it would be worth documenting how to do this in the README

Absolutely, will do

bkeepers:
If you're looking for alternatives in the short term, I think the implementation Probot has worked really well. It allows stopping pagination when a result is found, allows accumulating results into an array, and mostly avoids magic.

gr2m:
I’m thinking of exposing a `.paginate()` function on each endpoint method that paginates. That would make it less magic and would avoid side effects

```js
const results = []
for await (const result of github.issues.getForRepo.paginate(options)) {
  results.push(...result.data)
}
```

> If you're looking for alternatives in the short term, I think the implementation Probot has worked really well. It allows stopping pagination when a result is found, allows accumulating results into an array, and mostly avoids magic.

I think I’d implement that as well, as a default plugin

gr2m:
I’ll get to this shortly

gr2m:
@bkeepers this is ready now, I added examples to paginate until I find the thing I'm looking for
and map a field from the results into a single resulting array

I think I’ll deprecate the previous pagination methods to see if people use them at all. Any thoughts?

gr2m:
with v16 all methods will have the [`.endpoint()`](https://github.com/octokit/endpoint.js#readme) method, e.g. `github.issues.getForRepo.endpoint(options)`

So now I tend to do 

```js
for await (const result of github.paginate(github.issues.getForRepo.endpoint(options))) {
  results.push(...result.data)
}
```

instead of 

```js
for await (const result of github.issues.getForRepo.paginate(options)) {
  results.push(...result.data)
}
```

The `.paginate` method would then just be a slight wrapper around `.request`. And we could add an optional callback function as 2nd argument. If it’s passed, no asyncIterator is returned but instead would work similar to [Probot’s `context.github.paginate`](https://probot.github.io/docs/pagination/)

```js
const issueTitles = await github.paginate(
  github.issues.getForRepo.endpoint({owner, repo}),
  response => response.data.issues.title
)
```

But I’m not sure if we really need that 🤔 I think it’s only good for mapping, it’s not useful for [early exits like shown in Probot’s docs](https://probot.github.io/docs/pagination/#early-exit), the example showed there does not prevent more requests from being sent.

gr2m:
Sorry not sorry for thinking out loud here :) 

I don’t like the idea of `octokit.paginate()` returning an async iterator depending on wether a 2nd argument is passed or not, see my comment above.

Here is what I’m thinking right now:

```js
const options = github.issues.getForRepo.endpoint({owner, repo})

// returns array of all issues across all pages
const issues = await github.paginate(options)

// map results
const issueTitles = await github.paginate(options, response => response.data.issues.title)

// async iterator
for await (const response of github.paginate.iterator(options) {
  // do whatever you want with each response, break out of loop, etc.
}
```

gr2m:
What’s cool is that if folks would like to implement pagination for all the methods, they could do this in v16

```js
github.hook.wrap('request', (request, options) => {
  if (!options.request.paginate) {
    return request(options)
  }

  delete options.request.paginate
  return github.paginate(options)
})

const repos = await github.issues.getForRepo({owner, repo, request: {paginate: true}})
```

Issue #899, opened 2018-05-28
Allow to set default parameters / headers / authentication


gr2m:
This is somewhat related to https://github.com/octokit/rest.js/issues/898

As a developer of GitHub Applications, I’d like to get an API with presets for the current installation. The API would use right authentication and would have `owner` and `repo` parameters preset.

Something like this

```js
const {data: {token}} = await client.apps.createInstallationToken({installation_id})
const octokitRestRepo = client.scope({
  headers: {
    authorization: `Bearer ${token}`
  },
  owner: 'octokit'
  repo: 'rest.js'
})
const {data: {issues}} = octokitRestRepo.issues.getForRepo()
```

Similar to [`octokat`’s `.repos()` function](https://github.com/philschatz/octokat.js#overview).

Please comment if you have more use cases that we should take into consideration, or if you have any questions :)

JasonEtco:
So excited for this 😍I have a quick question: you mention that the preset values will be taken from the installation. That makes sense for Octokit in general, but as you know in the Probot world the defaults (`owner/repo`) are already in the webhook payload (which is especially helpful when the installation isn't for a specific repo).

I wonder how we could balance that? I see here that you've created a `.scope()` method;  would that be the right avenue to tack on the presets from somewhere else (like a webhook payload)?

bkeepers:
I love this.

It would be awesome if Probot could create a scoped client that has any variable that might be useful in responding to a webhook event (like `number` for something that looks issue-ish, `sha` for things that are commit-ish, etc), and Octokit would only pluck those values out if the endpoint called accepted those params.

gr2m:
I’m still wondering if the way defaults are defaults are set described in this issue a problem for typescript, as required parameters are not passed to the methods? Would love help by someone with more Typescript experience.

If it is, then we could consider changing the way way the REST API methods work to something more like https://github.com/philschatz/octokat.js/. I think that might work well with a scoped client passed to webhook handlers, too. E.g. instead of 

```js
client.issues.createComment({owner, repo, number, body})
```
it could be

```js
client.repo({owner, repo}).issue({number}).createComment({body})
```

Then when implementing webhooks in `probot` or a future all-batteries-included `octokit` we could pass in a pre-scoped `issue` client to an [issues event](https://developer.github.com/v3/activity/events/types/#issuesevent) handler

```js
webhooks.on('issues.opened', async ({issue}) => {
  await issue.createComment({body: 'Hello, world'})
})
```

[...]

Issue #889, opened 2018-05-24
Rename routes to align better with docs on https://developer.github.com/v3/


gr2m:
We are putting a lot of effort into making `@octokit/rest` a natural extension of the REST API docs. The goal is that when new REST routes are released in future they will become available in `@octokit/rest` immediately. 

The way the methods are called is being standardized now so they become more predictable. The downside of it is that we will need to rename a lot of method names. The current stat is 228 out of 474 methods will be renamed. 

In order to make the transition as painless as possible we will make sure to deprecate all methods that have been renamed. So none of your code will be broken and in order to upgrade to the next major release all you’ll need to do is to get rid of all the deprecation notices in your code. v16 won’t have any breaking changes besides the removal of deprecations.

The new method names are not set in stone yet. We are having a wider discussion with GitHub’s Documentation team as well as other Octokit maintainers. If you have any input on the name changes, now is the time :) I’ll take all your input to our upcoming octokit summit where we’ll hopefully make a decision on the method naming.

### How it works

[octokit/routes](https://github.com/octokit/routes) creates JSON files from the current documentation at https://developer.github.com/v3/. It runs a nightly cron job on Travis and each time it finds a change it creates a pull request ([example](https://github.com/octokit/routes/pull/147)). When merged, a new version gets released automatically and a pull request gets created for `@octokit/rest` ([example](https://github.com/octokit/rest.js/pull/876)). When that gets merged a new version of `@octokit/rest` gets released automatically. So we review the changes just to be sure, but the rest is automated.

The method names are based on the new `id-name` property introduced in https://github.com/octokit/routes/pull/158. It's calculated based on the section titles and the routes. As titles might change in future we will make sure that the titles themselves follow certain rules and are validated with a custom linter. That way everyone benefits.

Some of these rules are

1. If the response is paginated, start the title with “List …”
2. If the response returns an array but is not paginated, start the title with “Get all …” or just “Get …”, depending on wether all entities are returned or just a subset, like ["Get top referrers"](https://developer.github.com/v3/repos/traffic/#list-referrers).
3. If the response has no body (204), start the title with “Check …”
4. If the endpoint returns data for the currently authenticated user, don’t refer to them as “a user’s …”. That term is used in titles for endpoints that require a :username parameter.

### List of changed routes as of today (<a id="list-of-changes" href="#list-of-changes">#</a>)

moved to https://github.com/octokit/rest.js/blob/889/generate-routes-for-real/scripts/tmp-route-changes.md

[...]

PR #914, opened 2018-06-13
WIP User Migrations

sgvictorino:
I am making a first attempt at building user migrations into this library. If anyone would be willing to review the code and spot bugs or typos so this feature can possibly be merged in the future, that would be great.
Thanks in advance!

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -3842,7 +3842,7 @@
       },
       "url": "/repos/:owner/:repo/import"
     },
-    "deleteMigrationArchive": {
+    "deleteOrganizationMigrationArchive": {
       "headers": {
         "accept": "application/vnd.github.wyandotte-preview+json"
       },
@@ -3921,7 +3921,7 @@
       },
       "url": "/repos/:owner/:repo/import/large_files"
     },
-    "getMigrationArchiveLink": {
+    "getOrganizationMigrationArchiveLink": {
       "headers": {
         "accept": "application/vnd.github.wyandotte-preview+json"
       },
@@ -3942,7 +3942,7 @@
       },
       "url": "/orgs/:org/migrations/:migration_id/archive"
     },
-    "getMigrationStatus": {
+    "getOrganizationMigrationStatus": {
       "headers": {
         "accept": "application/vnd.github.wyandotte-preview+json"
       },
@@ -3963,7 +3963,7 @@
       },
       "url": "/orgs/:org/migrations/:migration_id"
     },
-    "getMigrations": {
+    "getOrganizationMigrations": {
       "headers": {
         "accept": "application/vnd.github.wyandotte-preview+json"
       },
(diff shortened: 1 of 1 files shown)


gr2m:
Hey @bitsol thanks for starting that. The routes.json files are updated automatically now using https://github.com/octokit/routes, but new routes need to be added manually still. Let me have a look into it

sgvictorino:
@gr2m I wish I had seen octokit/routes before. Is this PR still useful? With https://github.com/octokit/routes/commit/7f5a3f15e4e7c717a2b5f5d81fca3d9247a27c48, shouldn't routes.json have the user migration routes already? I thought I was simply adding a feature and clarifying route names, but that was already done?

gr2m:
@bitsol the PR is great, thanks! I totally missed the fact that we are lacking the User Migration APIs. Based on that I found some problems with the API Docs on migrations for both User & Organization Migrations, to help better distinguish the method names that will be are derived from the section titles in the v3 REST API docs. So your input was super valuable 👍 I can take it from here

If this is an urgent blocker for you right now? Or can it wait another few days?

sgvictorino:
Thank you so much!
The PR doesn't have to be reviewed and merged today or even tomorrow, as I could always point `npm` at my branch if absolutely necessary, but it would be nice to figure out in what ways these route/function names might evolve before that.

gr2m:
> it would be nice to figure out in what ways these route/function names might evolve before that

I was thinking about the naming yesterday. Looking at the conventions of the other method names (see #889) I’m tending towards

- `octokit.migrations.startForOrg()` for the organization (https://developer.github.com/v3/migrations/orgs/#start-a-migration)
- `octokit.migrations.startForUser()` for the currently authenticated user

etc.

What do you think?

sgvictorino:
@gr2m those names look good.

bitsol:
@gr2m am I stretching the idea too far when I name a method this: `unlockRepoLockedForMigrationForOrg`? 😜

bitsol:
Is that out of the scope of this PR, anyway? It looks like #889 will have all these changes.

sgvictorino:
@gr2m is this out of the scope of the PR?
I ask because it looks like what's currently `migrations.unlockRepoLockedForMigration` would have to be renamed to `migrations.unlockRepo`, probably as part of #899, to prevent names like this: `unlockRepoLockedForMigrationForOrg`. 😄 
In other words, what changes should I make before #914 (this PR) can be merged, while making sure to contribute to the future of this library?
Thanks again!

gr2m:
I’ll look more into it today and get back to you

sgvictorino:
@gr2m why didn't the routes update automatically though?

gr2m:
> why didn't the routes update automatically though?

The lib/routes.json file is not generated from ocotkit/routes, instead it’s matching the two based on the http verb + path. So if there are changes in the parameters or in the route description then lib/routes.json get updated automatically, but if there are entirely new routes nothing happens.

But that’s what https://github.com/octokit/rest.js/issues/889 is about, all these changes are coming from the method name being generated, too. So once we have that it should work all automatically, including adding new methods

Issue #916, opened 2018-06-13
Downloading a tarball of a migration archive


sgvictorino:
@gr2m, in response to https://github.com/octokit/rest.js/issues/881#issuecomment-397080133, below I have written a little user story for myself.
Using my #914 branch, I wanted to fetch the download link (to later write into a file) for a user migration in the "exported" state.
Executing the following code (I have authenticated already)
```javascript
octokit.migrations.getUserMigrationArchiveLink(
    {
      migration_id: id
    }, (error, result) => {console.log(result)})
```
displays
```
{ data: <Buffer 1f 8b 08 00 67 58 21 5b 00 03 ec 5b 07 5c 14 d7 d6 c7 04 23 58 10 13 7b f4 39 ac 1d d8 dd 29 db 51 a3 62 89 82 e2 13 44 05 14 67 77 87 dd 81 dd 99 75 ... >,
  status: 200,
  headers:
   { 'accept-ranges': 'bytes',
     'cache-control': 'max-age=2592000',
... 
```
while I expected a URL of this format to be included within `result`: `https://s3.amazonaws.com/github-cloud/migration/...`

How can I successfully fetch this link?

gr2m:
Can you elaborate on why you need the link? The `data` is the content of the archive, you can write it directly into a file

bitsol:
@gr2m Oh, so that's what `data` is. I was confused 😕 .

sgvictorino:
@gr2m Oh, so that's what `data` is. I was confused 😕 . Thanks for clearing that up.

For anyone who's wondering, the intended behavior can be achieved by following the second method of [this tutorial](http://stackabuse.com/writing-to-files-in-node-js/):
```js
fs.open(destination_path, "w", function(err, fd) {
            if (err) {
              throw "could not open file: " + err;
            }
            var buffer = result.data;
            fs.write(fd, buffer, 0, buffer.length, null, function(err) {
              if (err) throw "error writing file: " + err;
              fs.close(fd, function() {
                console.log("wrote the file successfully");
              });
            });
          });
```
where `result` is simply the response from the library's (at PR #914) `octokit.migrations.getUserMigrationArchiveLink` function, `destination_path` is a self-explanatory argument, and `fs` is the Node module that can be required with `require("fs")`.

I'm closing this issue now.

gr2m:
[...]

PR #921, opened 2018-06-16
Greenkeeper/@octokit/routes 9.0.2

gr2m:
closes https://github.com/octokit/rest.js/pull/914, closes https://github.com/octokit/rest.js/pull/920

ping @bitsol this PR introduces the missing methods for user migrations. Could you double check if they work for you?

The methods are

- `octokit.migrations.deleteArchiveForAuthenticatedUser()`
- `octokit.migrations.getArchiveForAuthenticatedUser()`
- `octokit.migrations.getStatusForAuthenticatedUser()`
- `octokit.migrations.listForAuthenticatedUser()`
- `octokit.migrations.startForAuthenticatedUser()`
- `octokit.migrations.unlockRepoForAuthenticatedUser()`

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -3843,6 +3843,9 @@
       "url": "/repos/:owner/:repo/import"
     },
     "deleteArchiveForAuthenticatedUser": {
+      "headers": {
+        "accept": "application/vnd.github.wyandotte-preview+json"
+      },
       "method": "DELETE",
       "params": {
         "migration_id": {
@@ -3874,6 +3877,9 @@
       "url": "/orgs/:org/migrations/:migration_id/archive"
     },
     "getArchiveForAuthenticatedUser": {
+      "headers": {
+        "accept": "application/vnd.github.wyandotte-preview+json"
+      },
       "method": "GET",
       "params": {
         "migration_id": {
@@ -4003,6 +4009,9 @@
       "url": "/orgs/:org/migrations"
     },
     "getStatusForAuthenticatedUser": {
+      "headers": {
+        "accept": "application/vnd.github.wyandotte-preview+json"
+      },
       "method": "GET",
       "params": {
         "migration_id": {
@@ -4013,6 +4022,9 @@
       "url": "/user/migrations/:migration_id"
     },
     "listForAuthenticatedUser": {
+      "headers": {
+        "accept": "application/vnd.github.wyandotte-preview+json"
+      },
       "method": "GET",
       "params": {
         "page": {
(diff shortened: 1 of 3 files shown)


sgvictorino:
Were there any issues with my #914 PR?
I also think #916 is a blocker for testing `octokit.migrations.getArchiveForAuthenticatedUser()`, but I'll review what I can when the time comes. 😃

gr2m:
You could try `octokit.migrations.getArchiveForAuthenticatedUser()` with a smaller archive maybe so it won’t hit the memory limit? 

#914 was great, it’s just that I most of the changes to lib/routes.json is happening automatically. All that’s needed is to add a new key to one of the scopes which will become the method name, the method and the url. Then run `npm run generate-routes` and it will add the rest automatically. So it was easier for me to do it that way. The new method names are also based on the `idName` properties from https://github.com/octokit/routes (https://github.com/octokit/routes/blob/cab6f02970cd1d953533bbd753520be7810ee558/index.json#L6626-L6765) so hopefully they won’t need to be changed again

sgvictorino (changes requested):
All the migration routes need to include "application/vnd.github.wyandotte-preview+json" in the "Accept" header.
See how it's done with the organization migrations: https://github.com/octokit/rest.js/blob/8225347b8d99ef4908bbb7e6f5e08f458c1b48a3/lib/routes.json#L4140

sgvictorino:
@gr2m 
Following #923, I've tested and can approve the changes to:

- [x] `octokit.migrations.deleteArchiveForAuthenticatedUser()`
- [x] `octokit.migrations.getArchiveForAuthenticatedUser()`
- [x] `octokit.migrations.getStatusForAuthenticatedUser()`
- [x] `octokit.migrations.listForAuthenticatedUser()`
- [x] `octokit.migrations.startForAuthenticatedUser()`
- [ ] `octokit.migrations.unlockRepoForAuthenticatedUser()` - This keeps returning 404s, even after putting it in a callback from `octokit.migrations.startForAuthenticatedUser()`. I have no idea what's going on. 😕 

Hope this helped!

sgvictorino:
I still have the same error.
Here's a little walkthrough. It doesn't show me changing the code, but the way the outputs evolve reflects that: https://asciinema.org/a/60v8qI9woD9bRmyY8ox66hNyI?t=10
*Edit: I think I locked almost all of my repos while testing this.* 😜

gr2m:
I’ll go ahead and merge this so we get the new APIs in, here is a follow up issue for the problem with `octokit.migrations.unlockRepoForAuthenticatedUser()`: #https://github.com/octokit/rest.js/issues/925

PR #923, opened 2018-06-16
proposed fix for #921: Define Accept header for all migration routes

sgvictorino:
Please see https://developer.github.com/v3/previews/#migrations and https://github.com/octokit/rest.js/pull/921#pullrequestreview-129370203.


gr2m (approved):
Great catch! Thanks for the PR 👍

PR #945, opened 2018-07-10
docs(examples): update removeAssigneesFromIssue example

### LINKED ISSUE #943: Removing assignees from issues is broken
It looks like removing assignees from issues using the `octokit.issues.removeAssigneesFromIssue()` method is broken on the current stable release. 

This is how I'm calling the method to an issue with only one assignee before the request:
```js
const r = await this.issues.removeAssigneesFromIssue({ // this represents the octokit client
    owner: 'synicalsyntax', repo: 'web', number: 23, body: 'synicalsyntax'
});
console.log(r.data.assignees.length); // returns 1
```

Debug logs for this request:
```
  octokit:rest REQUEST: { timeout: 0,
  agent: undefined,
  method: 'delete',
  url:
   'https://api.github.com/repos/synicalsyntax/web/issues/23/assignees',
  headers:
   { accept: 'application/vnd.github.v3+json',
     'user-agent': 'octokit/rest.js v15.9.4',
     authorization: '[REDACTED]' },
  body: { body: '{"assignees":"synicalsyntax"}' } }
```

The issue is that the`body` value is wrong (it should be `{"assignees":"synicalsyntax"}` instead), so the response returns a 200 OK code without actually removing the assignee from the issue.

Even though the GitHub API is at fault for not rejecting the request, the request body should be modified to have the correct value to resolve this bug.

synicalsyntax:
Current documentation example incorrectly describes the `assignees` parameter; this commit corrects it.

Fixes #943

### DIFF
--- a/examples/removeAssigneesFromIssue.js
+++ b/examples/removeAssigneesFromIssue.js
@@ -11,5 +11,5 @@ octokit.issues.removeAssigneesFromIssue({
   owner: 'octokit',
   repo: 'rest.js',
   number: '4',
-  body: { 'assignees': ['first9890'] }
+  assignees: ['first9890']
 })


gr2m (approved):
great, thank you!

PR #956, opened 2018-07-20
Add preview headers needed for the settings app

jwsloan:


gr2m:
Looks like the tests are failing due to fixtures from https://github.com/octokit/fixtures, would you like to have a look at it?

jwsloan:
Thanks for pointing that out, @gr2m! I forgot to come back and verify all the checks.

gr2m:
I think the `symmetra-preview` accept header is not required for the APIs to work. When you need it, you can add it to the request options

```js
const labels = ['foo', 'bar']
const headers = {
  accept: 'application/vnd.github.symmetra-preview+json'
}
client.issues.addLabels((owner, repo, labels, headers})
```

Maybe we should add that example to the README to make it clear?

jwsloan:
I'd be fine with the app passing in any preview headers it needs. I just couldn't figure out to do it. Can I pass headers at the end of any api call?

gr2m:
> Can I pass headers at the end of any api call?

Yes

I want to stay close to how the v3 APIs work, so I don’t want to pass in all existing preview headers to all requests but instead let the users do it themselves. The only exception is if an API does not work without the preview header, we agreed that in these cases we want to set the header by default

jwsloan:
@gr2m in the case of `updateBranchProtection`, I'd say it does not work without the preview headers, as the teams functionality within branch protection is not working at all without it. Maybe I could limit my PR to just that endpoint?

gr2m:
> @gr2m in the case of updateBranchProtection, I'd say it does not work without the preview headers, as the teams functionality within branch protection is not working at all without it. Maybe I could limit my PR to just that endpoint?

that’t be great, thanks

gr2m:
🔔 ping @jwsloan

jwsloan:
Thanks for the ping, @gr2m. I decided to try tackling this[ in the Settings app itself](https://github.com/probot/settings/pull/98). I thought it might be more straightforward. Sorry I didn't comment/close this one. 

In doing that work, it appears as though some of the endpoints needed for the Teams plugin are still not accessible to apps. :( Which makes me think we should just remove that plugin until it actually can function correctly. 

Do you agree that we should drop this and just do the work in the Settings app?

PR #957, opened 2018-07-23
chore(package): update @octokit/routes to version 11.4.0

gr2m:
Closes #947, closes #977


lib/routes.json
```diff
@@ -1539,7 +1535,9 @@
             "cancelled",
             "timed_out",
             "action_required",
-            "details_url"
+            "conclusion",
+            "status",
+            "completed"
```
gr2m, line 1540: see https://github.com/octokit/routes/issues/199

gr2m:
it shouldn’t break codes that use `head_branch` right now, we only removed the error being returned if it’s missing

PR #964, opened 2018-08-04
feat: allow to escape a + in a query

tleb:
Previously, cases such as

```javascript
octokit.search.repos({ q: 'language:"C++"' })
```
Got an error with the message: `None of the search qualifiers apply to this search type`. It was because [lib/endpoint/add-query-parameters.js](/octokit/rest.js/blob/master/lib/endpoint/add-query-parameters.js) treated the `q` parameter differently, by splitting it at each  `+`. At first, I started fixing this by escaping every `+` in quotes. But this felt too specific to my use case and complicated for no reason, so I preferred introducing escaping through the `\` character. Now:

```javascript
octokit.search.repos({ q: 'language:"C\\+\\+"' })
```

Works as intended.

Meaning: now, when you put data that might have a `+` in it into a query, you can replace it by `s.replace(/\+/g, `\\+')` whereas before it was just broken.

Do you guys have a better implementation idea? Another syntax?

Great lib!

### DIFF
--- a/lib/endpoint/add-query-parameters.js
+++ b/lib/endpoint/add-query-parameters.js
@@ -11,7 +11,7 @@ function addQueryParameters (url, parameters) {
   return url + separator + names
     .map(name => {
       if (name === 'q') {
-        return 'q=' + parameters.q.split('+')
+        return 'q=' + escapedSplit(parameters.q, '+')
           .map(encodeURIComponent)
           .join('+')
       }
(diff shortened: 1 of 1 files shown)


gr2m:
@tleb apologies for the late response and thank you for the pull request! I'm not 100% sure for the best solution here, I’ll need some more time to think about it. I’m catching up after parental leave and will get back to it soon

tleb:
Another idea: couldn't `querystring` be used? And what does the + syntax permit?

No worries, enjoy yourself 🙂

PR #977, opened 2018-08-14
Remove references to 'head_branch' from Checks API

dbolkensteyn:
I did not find any mention of `head_branch` in the Checks API documentation:
* https://developer.github.com/v3/checks/suites/ - except in the response examples
* https://developer.github.com/v3/checks/runs/

However, on the archived version of the page from the 8th of May, it does appear as a required parameter: https://web.archive.org/web/20180508231639/https://developer.github.com/v3/checks/suites/

So my guess is that this parameter was formerly required but has since been removed. Therefore, it should be safe to remove it as well from this client of the API.

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -1135,10 +1135,6 @@
         "external_id": {
           "type": "string"
         },
-        "head_branch": {
-          "required": true,
-          "type": "string"
-        },
         "head_sha": {
           "required": true,
           "type": "string"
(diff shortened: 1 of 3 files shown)


gr2m:
Thanks Dinesh! The `routes.json` file is automatically generated based on https://github.com/octokit/routes. We are a bit behind with updating to the latest, but once we do it will be resolved (via https://github.com/octokit/rest.js/pull/957).

We’ll still need to adopt the tests so thanks for that! I’ll leave your PR open as a reminder :)

PR #987, opened 2018-08-22
Repository Pre-receive Hooks for Enterprise

### LINKED ISSUE #986: Pre-receive hooks endpoints missing
The API and `routes.json` contain methods for administration of pre-receive hooks defined in:
https://developer.github.com/enterprise/2.14/v3/enterprise-admin/pre_receive_hooks/

But the repository settings for pre-receive hooks defined in the following spec are missing completely:
https://developer.github.com/enterprise/2.14/v3/repos/pre_receive_hooks/

Eddman:
Closes #986.

Added all methods for repository pre-receive hooks for GH Enterprise.

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -1724,6 +1724,27 @@
       },
       "url": "/admin/pre_receive_hooks/:id"
     },
+    "deleteRepoPreReceiveHook": {
+      "headers": {
+        "accept": "application/vnd.github.eye-scream-preview"
+      },
+      "method": "DELETE",
+      "params": {
+        "owner": {
+          "required": true,
+          "type": "string"
+        },
+        "pre_receive_hook_id": {
+          "required": true,
+          "type": "string"
+        },
+        "repo": {
+          "required": true,
+          "type": "string"
+        }
+      },
+      "url": "/repos/:owner/:repo/pre-receive-hooks/:pre_receive_hook_id"
+    },
     "editPreReceiveEnvironment": {
       "method": "PATCH",
       "params": {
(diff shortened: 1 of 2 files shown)


lib/routes.json
```diff
@@ -10240,6 +10240,90 @@
...
       },
       "url": ":url"
+    },
+    "getPreReceiveHook": {
+      "headers": {
+        "accept": "application/vnd.github.eye-scream-preview"
+      },
+      "method": "GET",
+      "params": {
+        "id": {
```
gr2m: please use `pre_receive_hook_id` instead of `id` as parameter to align with the documentation at https://developer.github.com/enterprise/2.14/v3/repos/pre_receive_hooks/. Update in `params: {...}` objects and the `url` string

Eddman:
@gr2m done.

gr2m:
CI currently fails because of the way the routes.json file is automatically updated based on octokit/routes. I’m looking into it

gr2m:
I move the routes to the `enterprise` namespace to make it work with the current way the `routes.json` file is updated (the enterprise namespace is left as-is). I remember talking to the GitHub docs team about how the two different pre-receive hooks pages are confusing ...

So the new endpoints added are

- `client.enterprise.deleteRepoPreReceiveHook`
- `client.enterprise.getRepoPreReceiveHook`
- `client.enterprise.getRepoPreReceiveHooks`
- `client.enterprise.updateRepoPreReceiveHook`

They are not documented at https://octokit.github.io/rest.js/ because of the way it is all setup right now. But I hope it will unblock the problem for you for now?

gr2m:
Okay it fails now due to the Node 4 tests. I think this is unrelated and we will officially drop Node 4 support in the upcoming version anyway. I’m working on the v16 release but there is some more work left, can you install from your fork for the time being?

Eddman:
For the time being I did a dirty workaround by replacing the `routes.json` before starting my client :rofl: 

However would be nice to have a proper solution soon....

Jenisonn:
http://p.codekk.com/



Sent from my Samsung Galaxy smartphone.

gr2m:
I’ll release it shortly before v16, once all the other work is done

Eddman:
@gr2m What is the status of this one and of this one: #997? I'd like to remove my workarounds... :smirk:

gr2m:
#997 should get resolved via #1043, but in the way that the methods that don’t have type definitions will be removed ¯\\_(ツ)\_/¯ The `.getById()` methods are not documented so we will ask folks to use `client.request()` for the time being, I know that GitHub folks are discussing to make these APIs public.

It also removed all enterprise-only methods which I want to bring back with an official plugin, this issue is a good one to watch for updates: https://github.com/octokit/routes/issues/254. Ones we successfully scrape the enterprise API docs, the plugin should be easy to build.

gr2m:
we will remove all enterprise and SCIM routes from `@octokit/routes` in v16, but provide official plugins that bring them back when needed. The one for SCIM is ready: https://github.com/octokit/plugin-scim.js. The one for the enterprise routes will need a little more work.

I’m sorry for the delay, I fear you’ll have to keep the workaround for a while longer.

PR #993, opened 2018-08-28
chore(package): Move @gimenete/type-writer to devDependencies

### LINKED ISSUE #1004: move "@gimenete/type-writer" to devDependencies
follow up for https://github.com/octokit/rest.js/pull/948#discussion_r213400451

bkeepers:
I asked about this in https://github.com/octokit/rest.js/pull/948#discussion_r213400451, but thought I'd open a PR to discuss. It looks like this is only used in the build process, so I don't think it needs to be a runtime dependency.

cc @gimenete @gr2m 

closes https://github.com/octokit/rest.js/issues/1004

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "node": ">=4"
   },
   "dependencies": {
-    "@gimenete/type-writer": "^0.1.3",
     "before-after-hook": "^1.1.0",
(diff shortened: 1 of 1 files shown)


gimenete:
You are totally right. Your change looks good to me 👍

gr2m (approved):

PR #998, opened 2018-08-30
Add authentication as a GitHub app

BPScott:
Allows you to pass in an `appId` and `privateKey` to the authenticate
method to authenticate as an app

Allows you to pass in an `appId`, `privateKey` and `installationId` to
the authenticate method to authenticate as an installation of an app.

When passing `installationId` it will request an access token for that
installation, cache the token and use it for subsequent requests.
Shortly before the token for the installation expires a new token shall
be automatically requested.

---

Here's the implementation. I've not wrote tests just yet in case people have feedback about the approach.

This solves the usecase for private apps where there is only ever one installation.
For public apps where the consumer of octokit needs to act on behalf of multiple installations they shall need to create multiple instances of octokit - one for each installation.

You can see this in action:

Save the following examples (replacing appId and installationId as appropriate) as `testing.js` within the root of the octokit repo and run them with: `DEBUG=octokit:rest* node ./testing.js` to see API requests 

### Authenticating as an application:

```js
const fs = require('fs')

const keyPath = 'filepath_to_an_application_private_key'
const octokit = require('.')({debug: true})
const appId = 14631 // Change to your app ID 
const installationId = 312666 // Change to your installation ID

octokit.authenticate({
  type: 'app',
  appId: appId,
  privateKey: fs.readFileSync(keyPath, 'utf8')
})

octokit.apps.getInstallations().then(console.log)
```

### Authenticating as an installation

```js
const fs = require('fs')

const keyPath = 'filepath_to_an_application_private_key'
const octokit = require('.')({debug: true})
const appId = 14631 // Change to your app ID 
const installationId = 312666 // Change to your installation ID

octokit.authenticate({
  type: 'app',
  appId: appId,
  installationId: installationId,
  privateKey: fs.readFileSync(keyPath, 'utf8')
})

octokit.apps.getInstallationRepositories().then((response) => {
  console.log(response);

  octokit.apps.getInstallationRepositories().then((response) => {
    console.log('A second request, note it reusing installation tokens rather than requesting them every time');
    console.log(response);
  })
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,23 @@ octokit.authenticate({
   type: 'app',
   token: 'secrettoken123'
 })
+
+// GitHub app, authenticating as the application
+// Allowing access to the endpoints marked as requiring a JWT at https://developer.github.com/v3/apps/
+octokit.authenticate({
+  type: 'app',
+  appId: 'your_app_id',
+  privateKey: 'your_private_key'
+})
+
+// GitHub app, authenticating as an installation
+// Allowing access to the endpoints at https://developer.github.com/v3/apps/available-endpoints/
+octokit.authenticate({
+  type: 'app',
+  appId: 'your_app_id',
+  installationId: 'installation_id',
+  privateKey: 'your_private_key'
+})
 ```
 
 Note: `authenticate` is synchronous because it only sets the credentials
--- a/README.md
+++ b/README.md
@@ -141,13 +141,32 @@ octokit.authenticate({
   type: 'app',
   token: 'secrettoken123'
 })
+```
+
+Authenticating as a Github App is slightly more involved. It is up to you to
+provide a JSON Web Token, as this process differs between node and browsers.
+On node we recommend using the `jsonwebtoken` package.
+
+```javascript
+const jsonwebtoken = require('jsonwebtoken')
+
+function jwt (id, pem) {
+  const payload = {
+    iat: Math.floor(Date.now() / 1000), // Issued at time
+    exp: Math.floor(Date.now() / 1000) + 60, // JWT expiration time
+    iss: id // Integration's GitHub id
+  }
+
+  // Sign with RSA SHA256
+  return jsonwebtoken.sign(payload, pem, {algorithm: 'RS256'})
+}
 
 // GitHub app, authenticating as the application
 // Allowing access to the endpoints marked as requiring a JWT at https://developer.github.com/v3/apps/
 octokit.authenticate({
   type: 'app',
   appId: 'your_app_id',
-  privateKey: 'your_private_key'
+  appTokenGenerator: (id) => jwt(id, 'your_private_key')
 })
 
 // GitHub app, authenticating as an installation
@@ -156,7 +175,7 @@ octokit.authenticate({
   type: 'app',
   appId: 'your_app_id',
   installationId: 'installation_id',
-  privateKey: 'your_private_key'
+  appTokenGenerator: (id) => jwt(id, 'your_private_key')
 })
 ```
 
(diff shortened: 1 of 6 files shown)


BPScott:
It looks like the majority of the bundle size increase is coming from jsonwebtoken. Upon closer inspection that package is all good for node but doesn't really work in browser land so it's a bit of a non-starter.

I'm not sure if the implementation of the LRU cache will work in browser land either. It'd be worth investigating alternative implementations.

I could rework this so that instead of passing in an `appId` and `private key` and generating the token internally I could pass a `tokenGenerator` function so it's up to the consumer to work out how to build a token. It would remove that dependency however it means the token generation is in the user's hands which isn't an ideal plug-and-play experience.

BPScott:
Gonna YOLO merge this once it gets approval, as I've proved it works locally with a my test APP/Installation Id, but there's no way to prove it works for the shopify app till it is deployed.

BPScott:
@gr2m & @bkeepers: Could I get a review and some feedback on this approach please?

I'm pretty happy that this solves the "private app" use case where there is only one installation of an application.

For the "public app" use case where a codebase has to authenticate itself as multiple installations I think it is best to say: an instance of octokit can only be authenticated by one method at a time, and that if a codebase needs to make requests as multiple installations then it has to maintain a set of octokit instances - one per installation.

gr2m:
Sorry for the delay, I’ll have a look at it

gr2m:
Sorry for the delay Ben, your PR looks great but I’m not sure if it’s worth merging it as is, we are working on https://github.com/octokit/app.js and https://github.com/octokit/octokit.js that will solve this properly. Your PR is certainly helpful to compare with our own implementations there.

Is this a blocker for you right now or could you work around it?

BPScott:
I've been able to work around this by creating my own plugin that handles authentication rather than using the built-in authentication plugin. It works for now but I was a little bit wary of it as you're currently claiming the plugin api isn't public and is subject to change.

gr2m:
I think the currently private plugin API won’t change. What will change is the way you define asynchronous hooks, but migrating will be easy. I’ll happily help you out if you get stuck upgrading to v16

wolfy1339:
I don't think this should be marked as stale as it's attached to a milestone

gr2m:
This won’t be implemented in `@octokit/rest`, but will be part of [`octokit`](https://github.com/octokit/octokit.js). I’ll leave the PR open as a reminder and also for folks to subscribe to that are interested in it

gr2m:
@BPScott  the plugin API is now public and [`@octokit/app`](https://github.com/octokit/app.js) is also released now that you could used in your plugin. Is it public? Maybe it might be useful to others, too?

gr2m:
Closing due to lack of activity. Follow https://github.com/octokit/octokit.js for updates on combining GitHub App authentication and REST API endpoint methods

BPScott:
Whoops, forgot about that.  Glad to hear the plugin API and `@octokit/app` are both released. I ended up moving away from node as part of that project for other reasons so I stopped paying attention to this PR.

There's not much value in me publishing this plugin given I'm not using it myself anymore and the long-term plan is to move to octokit.js. Anybody who needs it now can extract it from this PR.

gr2m:
fyi, once #1215 is released, you’ll be able to do this:

```js
const App = require('@octokit/app')
const Octokit = require('octokit')

const app = new App({ id: process.env.APP_ID, privateKey: process.env.PRIVATE_KEY })
const octokit = new Octokit({
  auth () {
    return app.getInstallationAccessToken({ process.env.INSTALLATION_ID })
  }
})
```

PR #999, opened 2018-08-31
Add the namespace to the generated return types

BPScott:
This ensures each response type is unique, as there are cases where the
same function name exists under multiple namespaces - e.g.
`octokit.issues.getForUser` and `octokit.users.getForUser` (along with
many other getForUser functions).

Prior to this commit a single type `GetForUserResponse` was created,
which was resulted in multiple functions using this one type, and
type being wrong for many of those functions.

Now multiple types are defined - `IssuesGetForUserResponse` and
`UsersGetForUserResponse` etc - one for each function-in-a-namespace and there is no
overlap.

//cc @gimenete and @gr2m

### DIFF
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -127,7 +127,7 @@ function generateTypes (languageName, templateFile, outputFile) {
       let responseType = 'Github.AnyResponse'
       if (entry[1].responses) {
         const typeName = 'Github.' + typeWriter.add(entry[1].responses.map(response => response.body || {}), {
-          rootTypeName: pascalcase(`${entry[0]}Response`)
+          rootTypeName: pascalcase(`${namespace}-${entry[0]}Response`)
         })
         responseType = 'Github.Response<' + typeName + '>'
       }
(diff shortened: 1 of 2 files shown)


gimenete:
Good catch. Regarding semver affecting this change: do we consider this a breaking change? It is something that the compiler will catch and does not affect business logic.

BPScott:
>  do we consider this a breaking change?

If you use the returned data from `users.getForUser` and update from 15.9.x to 15.10.0 your code will no longer compile as typescript will claim things like `response.data.login` do not exist, despite it being totally correct and working as expected. This is a bug fix as without it previously working code is uncompilable.

BPScott:
Rebased atop of master to dodge conflicts.

BPScott:
Yo @gr2m would it be possible to get this looked at? When updating to >15.10.0 I get typescript compilation errors on previously valid code like:

```ts
const user = await octokit.users.getForUser({username: 'octokit'})
console.log(user.data.login)
```

I can "fix" it by doing the following, but it's not ideal:

```ts
const user = await octokit.users.getForUser({username: 'octokit'})
console.log((user.data as any).login)
```

gr2m (approved):
Thanks 💐

gr2m:
It’s not a breaking change because it only affects the build step, not code run in production

PR #1002, opened 2018-09-02
fix(routes): update `check_run_id` and `check_suite_id` to integers (#1001)

### LINKED ISSUE #1001: Checks API routes have incorrect types for id's
TypeScript compiler fails when for interfaces with `check_run_id` and `check_suite_id`.
These are `strings` however in the docs and API response expects a `number`.

This is due to the routes.json file has them as `strings`, and they are converted to .d.ts files later.
Currently the only way around this is by using `// @ts-ignore` or marking `as any` inline.

update check body
https://github.com/octokit/rest.js/blob/master/lib/routes.json#L1523-L1525

eknowles:
This PR addresses the issue of incorrect interface's being generated from the routes (`scripts/routes-for-api-docs.json`).

`check_run_id` - `string` => `integer`
`check_suite_id` - `string` => `integer`

Closes #1001

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -1263,7 +1263,7 @@
       "params": {
         "check_run_id": {
           "required": true,
-          "type": "string"
+          "type": "integer"
         },
         "owner": {
           "required": true,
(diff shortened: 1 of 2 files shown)


gr2m:
see https://github.com/octokit/rest.js/issues/1001#issuecomment-420399692

Issue #997, opened 2018-08-29
TypeScript definition missing methods


Eddman:
In 15.10.0 after merge of #948 the TypeScript definitions are generated using `routes-for-api-docs.json` instead of `routes.json`. But the file is not complete and is missing multiple methods including eg. `repos.getById` or the whole `enterprise` section:
https://github.com/octokit/rest.js/pull/948/files#diff-70e044778205a59a14c885cb6804b03eR14

kkweon:
Some params are generated with wrong params. It's also related to this ?

For example, 
```ts
  export type IssuesReplaceAllLabelsParams = {
    owner: string;
    repo: string;
    number: number;
    // missing `labels: string[];`
  };
```

And here's the official [API doc](https://developer.github.com/v3/issues/labels/#replace-all-labels-for-an-issue)

BPScott:
`users.getForUser()` is also not getting the correct typehints for me after updating to 15.10.0.

You can reproduce this by adding the following to `test/typescript-validate.ts` then running `npm run validate:ts`:

```ts
const user = await octokit.users.getForUser({username: 'octokit'})
// Check Response
user.login
user.type
```

BPScott:
I believe this is two separate issues:

The removal of repos.getById was an explicit decision to remove undocumented API endpoints in 66c627fa. This occured several months ago, prior to the TypeScript definition updates in #948. It also affects 15.9.4 - the last version before 15.10.0 which included the the return type updates.

The removal of the enterprise section and  `IssuesReplaceAllLabelsParams` losing its labels attribute however IS related to the typescript definition updates.

You can recreate these issues by adding the following to the bottom of `test/typescript-validate.ts` and then running `rm -rf index.d.ts && npm run validate:ts`:

```
  await octokit.issues.addLabels({
    owner: 'octokit',
    repo: 'rest.js',
    number: 10,
    labels: ['label'],
  });

  await octokit.enterprise.stats({
    type: 'issues',
  });
```

/cc  @gimenete

kkweon:
How is `routes-for-api-docs.json` created? 

I suppose it's automatically generated based on the official doc.

Basically, https://developer.github.com/v3/issues/labels/#replace-all-labels-for-an-issue page doesn't clearly indicate **POST body** has to be `string[]`

and while `routes-for-api-docs.json` is generated, it doesn't take that into consideration.

So, the type definitions are missing the simple body property, and therefore, any other endpoints that takes something simple POST body would have the same issue.

[...]

PR #1007, opened 2018-09-04
Expose mapped parameters in routes-for-api-docs.json

BPScott:
Previously items with mapTo properties were not exposed in the API docs,
this means that when looking at the API docs you didn't know you had to
pass in a "labels" property to issues.addLabels.

This also means these properties will be exposed in the typescript
definitions, thus fixing @kkweon's issue in https://github.com/octokit/rest.js/issues/997#issuecomment-417406493

Compare https://octokit.github.io/rest.js/#api-Issues-addLabels with the following from running the docs locally after making this change, noting the extra "labels" field:

screen shot 2018-09-04 at 4 44 15 pm

### DIFF
--- a/scripts/generate-routes.js
+++ b/scripts/generate-routes.js
@@ -234,6 +234,12 @@ Object.keys(CURRENT_ROUTES).sort().forEach(scope => {
       if (currentParams[name].alias || currentParams[name].mapTo) {
         currentEndpoint.params[name] = currentParams[name]
       }
+
+      // Add the mapped name to the newEndpoint so that we expose the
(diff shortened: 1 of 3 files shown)


gr2m:
Fabolous PR, thank you Ben 💐 Will merge if it’s all green

gr2m:
ping  @kkweon can you confirm that it’s fixing your issue?

BPScott:
As a heads up, confirming in a separate project is going to be a pain until this is released - I tried changing my package.json to point to a branch of this repo in my typescript project but found that the type definitions were not present, as they're a build artifact rather than something checked into the repo.

PR #1008, opened 2018-09-05
Update createFromIssue

thom4parisot:
It seems more logical to run `pullRequests.createFromIssue()` with an issue number and because `title` was a mandatory field, it could never work without throwing an error like this

> ERROR event: {"message":"You may only provide an issue number or a title string, but not both.","documentation_url":"https://developer.github.com/v3/pulls/#create-a-pull-request"}

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -6104,6 +6104,10 @@
           "required": true,
           "type": "string"
         },
+        "issue": {
+          "required": true,
+          "type": "integer"
+        },
         "maintainer_can_modify": {
           "type": "boolean"
         },
@@ -6112,10 +6116,6 @@
           "type": "string"
         },
         "repo": {
-          "required": true,
-          "type": "string"
-        },
-        "title": {
           "required": true,
           "type": "string"
         }


BPScott:
Heeeeey it's Tom!

The code in `lib/routes.json` is generated from content in the [`@octokit/routes`](https://github.com/octokit/routes) repository, which in turn is generated from the Github API docs. In this case, the content at https://developer.github.com/v3/pulls/#create-a-pull-request

This PR would get overwritten the next time update from `@octokit/routes` are pulled in to this repo. The better fix for this issue would be to do something in `@octokit/routes` that handles the "alternative input" headings in the API docs.

thom4parisot:
Ah @BPScott, you here! So funny 🙂 

Thanks for the pointer, super helpful. From what I understand, octokit/routes#200 might be related.

In the meantime, I'll use the experimental `request` API.

Thank you for your help. I hope Vancouver treats you well 🙂

BPScott:
> @BPScott thank you for jumping in and helping out, very much appreciated!

Would have been better if I answered correctly though ;)

gr2m (approved):
thank you all 💝

Issue #1009, opened 2018-09-09
Some members of ReposUpdateBranchProtectionParams should be typed as optional


ianwremmel:
`required_status_checks` and `restrictions` can be `null`, according to current api behavior. `restrictions`, in particular, must be null for non-org projects.

gr2m:
The parameters are not optional, but can be set to null, as described in https://developer.github.com/v3/repos/branches/#parameters-1

I don’t see the error? Can you provide a code example, explain the error you get and the describe the expected result?

ianwremmel:
<img width="514" alt="screen shot 2018-09-11 at 10 39 18 pm" src="https://user-images.githubusercontent.com/1182361/45404402-94177800-b613-11e8-8617-dad1b658a22a.png">


Sorry, I was tired and wasn't as clear as I should have been. I'm referring to the generated types file. The typedef exported by the library doesn't allow for `required_status_checks` or `restrictions` to be `null`.

gr2m:
ping @gimenete :)

gimenete:
I've just created a PR that fixes it Check it out! :) #1024

ianwremmel:
awesome, thanks!

PR #1024, opened 2018-09-17
Fix nullable params in TypeScript

gimenete:
Closes https://github.com/octokit/rest.js/issues/1009

There were a total of 5 params that had `allowNull: true` in the routes file but that information was not being used when generating the type definitions file.


gr2m (approved):
thanks!

PR #1031, opened 2018-09-25
Using custom http-client for request sending

juliethefox:
Allows you to pass custom http client via user options.

```javascript
const got = require('got'); //for example

const github = new GitHubApi({
    httpClient: got,
    baseUrl: makeUrl()
});
```

We need the ability to pass our own module to make network calls. Used ```node-fetch``` unfortunately does not satisfy us, because in our case there is a complex logic of request retries in an unstable network environment, as well as own logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,10 @@ const octokit = require('@octokit/rest')({
   baseUrl: 'https://api.github.com',
 
   // Node only: advanced request options can be passed as http(s) agent
-  agent: undefined
+  agent: undefined,
+
+  // User function to make http calls can be passed
+  httpClient: undefined
 })
 ```
 
--- a/lib/parse-client-options.js
+++ b/lib/parse-client-options.js
@@ -65,6 +65,11 @@ function parseOptions (userOptions) {
       clientDefaults.baseUrl += '/' + userOptions.pathPrefix.replace(/(^[/]+|[/]+$)/g, '')
     }
   }
+
+  if (userOptions.httpClient) {
+    clientDefaults.request.client = userOptions.httpClient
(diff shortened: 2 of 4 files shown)


gr2m:
we will implement retries and expose an option for custom logging, but have no plans of using your own request library at this point.

We will also further modularize the octokit client so you will be able to only use the parts you need. For a preview, check out https://github.com/octokit/rest.js/tree/master/lib/endpoint. Just using the endpoint method which compiles the request options sounds like the best option in your case, it gives you full control of the actual request

gr2m:
let me know if you have more questions Julie, I’ll close it for the time being

gr2m:
> in our case there is a complex logic of request retries in an unstable network environment

I’m looking to build in retry logic and throttling  into `@octokit/rest` myself. Could you share what you have done in your case?

juliethefox:
> > in our case there is a complex logic of request retries in an unstable network environment
> 
> I’m looking to build in retry logic and throttling into `@octokit/rest` myself. Could you share what you have done in your case?

In our case we created package with with support for the following options:
- retries time & count of attempts: options retry count, retry total time, max retry time, max retry after 
- retriable status codes (by default - all codes except 405, 413, 414, 418, 431, 508)
- retriable network errors: https://github.com/sindresorhus/got/blob/25f18be291dd866a37f5153ca76d44151fcd702b/source/is-retry-on-network-error-allowed.js#L4-L11
- header `retry-after` support
- retries for POST (disabled by default, as it can cause side effects, but sometimes there is a need)

Also additional logging features implemented for linking request and it's retries

PR #1037, opened 2018-10-03
Update routes

### LINKED ISSUE #1001: Checks API routes have incorrect types for id's

gr2m:
fixes #1001

ping @eknowles can you check if it fixes your problem with `check_run_id` / `check_suite_id`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "@gimenete/type-writer": "^0.1.3",
     "@gr2m/node-fetch": "^2.0.0",
     "@octokit/fixtures-server": "^3.0.0",
-    "@octokit/routes": "12.0.0",
+    "@octokit/routes": "14.0.0",
     "@types/node": "^10.1.2",
     "apidoc": "^0.17.6",
     "bundlesize": "^0.17.0",
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -9,7 +9,7 @@
         },
         "thread_id": {
           "required": true,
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
     "@gimenete/type-writer": "^0.1.3",
     "@gr2m/node-fetch": "^2.0.0",
     "@octokit/fixtures-server": "^3.0.0",
-    "@octokit/routes": "12.0.0",
+    "@octokit/routes": "14.0.0",
```
hiimbex, line 53: Will this PR include the changes from https://github.com/octokit/routes/pull/267 based on this version bump?

gr2m: Yes :) Good point I’ll make sure to mention it in the release notes

gr2m: also I totally forgot that I still have to add the new endpoint to lib/routes.json as the file is not yet fully generated from octokit/routes, changes are only merged into it. But I’ve done that via 90aedbb now, so it’ll be in the next release. Thank you 🙏

PR #1043, opened 2018-10-06
[WIP] generate routes for real

### LINKED ISSUE #889: Rename routes to align better with docs on https://developer.github.com/v3/

gr2m:
this will eventually close #889. See summary of changes at https://github.com/octokit/rest.js/blob/889/generate-routes-for-real/scripts/tmp-route-changes.md

The `lib/routes.json` file is currently being automatically updated based on updates to [octokit/routes](https://github.com/octokit/routes). It is not replaced but instead changes are merged into the currentk `lib/routes.json` file in order to avoid breaking changes. That was an in-between solution. For example, that means that new routes will not automatically be added to `lib/routes.json`, we have to add them by hand.

This PR will track the progress on generating the `lib/routes.json` entirely from scratch and only merge in aliases to avoid breaking changes, so users can update their code based on the deprecation notices they see, as plenty of methods will be renamed to better align with the docs at https://developer.github.com/v3/, see https://github.com/octokit/rest.js/issues/889#list-of-changes for an overview

### TODOs

- [x] add aliases for methods that got renamed
- [x] re-generate method specifications from octokit/routes
  - [x] keep parameter aliases intact
  - [x] get required preview headers from octokit/routes (currently implemented with a workaround by reading the headers from `lib/routes.json`). Depends on https://github.com/octokit/routes/issues/146
  - [x] get mapTo for parameters that are sent in the root of the request body. The API team is working on removing parameters to be sent in the request body, e.g. see https://github.com/octokit/routes/pull/270/commits/429d4ef5a31b8cc845f2479015c433a970b75473#diff-ed2852752c2a2c008b3f3e257247ecabL7
  - [x] figure out enums with placeholders, depends on https://github.com/octokit/routes/issues/121
- [x] add new methods
- [x] incorrect method name (https://github.com/octokit/routes/issues/287)
  - [x] createCommentAlternative should be createCommentReply
  - [x] createFromIssueByPassingIssueNumber should be createFromIssue
- [x] missing aliases
  - [x] apps.getInstallationRepositories -> apps.listRepos
  - [x] apps.getInstallations -> apps.listInstallations
  - [x] users.addRepoToInstallation -> apps.addRepoToInstallation
  - [x] users.removeRepoFromInstallation -> apps.removeRepoFromInstallation
  - [x] misc.getEmojis -> emojis.get
  - [x] misc.getMeta -> meta.get
- [x] removed
  - [x] gitdata.getReferences (should not get removed, check script)
  - [x] gitdata.getTags (explain to use `gitdata.getReferences()` with 'tags/' as `namespace` parameter)
  - [x] missing in octokit/routes (https://github.com/octokit/routes/issues/50, 🔜 https://github.com/octokit/routes/pull/284)
    - [x] misc.getEmojis (emojis.get)
    - [x] misc.getMeta (meta.get)
  - [x] not documented -> explain workaround with client.request(`/`)
    - [x] repos.getById
    - [x] users.getById
  - [x] enterprise routes:
    - [x] users.demote (DELETE /users/:username/site_admin)
    - [x] users.promote (PUT /users/:username/site_admin)
    - [x] users.suspend (PUT /users/:username/suspended)
    - [x] users.unsuspend (DELETE /users/:username/suspended)
- [x] params
  - [x] author.email & author.name added in createFile,updateFile
  - [x] getDeploymentStatus,listDeploymentStatuses add invalid "id" property (reported to the docs team, problem is that the parameters table lists `id` params at [Get a single deployment status](https://developer.github.com/v3/repos/deployments/#get-a-single-deployment-status) and [List deployment statuses](https://developer.github.com/v3/repos/deployments/#list-deployment-statuses))
  - [x] get required Content-type header ([render raw markdown endpoint](https://developer.github.com/v3/markdown/#render-a-markdown-document-in-raw-mode) only)
  - [x] figure out header parameters (Content-Length & Content-Type for [upload a release asset](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) only)
- [x] remove all aliases / deprecated methods and parameters
- [x] do not copy preview headers from existing lib/routes.json
- [x] fully generate lib/routes.json from octokit/routes, without the need of reading out anything from lib/routes.json. That means in future aliases would need to live in octokit/routes

### Next

this PR needs to be split up, all changes that are not breaking go into v15, the rest goes into `next` / v16


gr2m:
closing in favor of #1094 and #1097

Issue #684, opened 2018-01-02
Parse JSON from GitHub error messages


bkeepers:
The `message` on errors thrown due to 4xx and 5xx errors from the GitHub API is currently a JSON string, such as (formatted for readability):

```json
{
  "message":"Invalid request.\n\nFor 'properties/name', false is not a string.",
  "documentation_url":"https://developer.github.com/v3/issues/labels/#create-a-label"
}
```

This adds a lot of noise when trying to read error logs. It'd be ideal if this we formatted in a more human-friendly way.

bkeepers:
Another example:

```json
{
  "message": "Validation Failed",
  "errors": [{
    "message":"The year 1918 in \"1918-01-27T22:55:01\" is outside the accepted range 1970 to 2970.",
    "resource":"Search",
    "field":"q",
    "code":"invalid"
  }],
  "documentation_url":"https://developer.github.com/v3/search/"
}

**Comment by gr2m** [comment]
Just so I understand correctly, instead loggin like this:

```
{"message":"Invalid request.\n\nFor 'properties/name', false is not a string.","documentation_url":"https://developer.github.com/v3/issues/labels/#create-a-label"}
```

You would like to have it log only the message, like so?

```
Invalid request.\n\nFor 'properties/name', false is not a string.
```

Or for the second example sth like 

```
Validation Failed: The year 1918 in "1918-01-27T22:55:01" is outside the accepted range 1970 to 2970.
```

?

**Comment by bkeepers** [comment]
Something like that, yeah. All the properties could still be set on the error object, so you'd have:

```
try {
  github.issues.search(params)
} catch (err) {
  console.log(err.message, err.documentation_url)
}
```

gr2m:
done via https://github.com/octokit/rest.js/pull/1052

PR #1052, opened 2018-10-15
v16

### LINKED ISSUE #601: support dry-run variable
When initializing the GitHubAPI client, there should be a property exposed which stubs out the actual request calls so it supports dry-run/log-only functionality. This would be great for use in testing situations.

### LINKED ISSUE #687: feature request: add the root endpoint
Hi maintainers of `octokit/node-github` 👋 

`https://api.github.com/` is the GitHub API v3's [root endpoint](https://developer.github.com/v3/#root-endpoint). I think it would be valuable to add this because it exists as a public resource but isn't currently implemented.

Here's my first pass:

https://github.com/francisfuzz/node-github/pull/1

Before I open a pull request against `octokit/node-github`, I have some questions for the team's consideration:

* If this were to be implemented, is it correct to write this test under the `examples/` directory, or somewhere else?
* Is it correct to use [Lisp case/kebab-case](https://en.wikipedia.org/wiki/Naming_convention_(programming)#Lisp) for `root-endpoint`? Would that be the best name to use or is there a better one to describe this?

Thanks ya'll ✌️

### LINKED ISSUE #688: Autopaginate

### LINKED ISSUE #766: remove flow type definitions
I made a quick poll (also on twitter) and it looks like nobody is using `octokit/rest.js` with [Flow](https://flow.org/)
https://github.com/octokit/rest.js/issues/620#issuecomment-363663120

I don’t even know if the current type definitions work. If you would like us to keep the flow type definitions please comment here, we'll have to figure out a way to make sure the type definitions don’t break as we make new releases using a test script.

Otherwise I'll return the flow type definitions with the upcoming milestone

### LINKED ISSUE #777: 💡💭 make octokit.request return response options
follow up for https://github.com/octokit/rest.js/issues/742#issuecomment-368351394

> Instead of having it figure out what to return based on response headers, I'll do what [r2](https://github.com/mikeal/r2) does and return an object that will give folks direct access to the response or get json, text, etc responses as they want, sth like this
> 
> ```js
> const jsonResponse = octokit.request({method: 'get', url: '/'}).json
> const textResponse = octokit.request({method: 'get', url: '/'}).text
> // etc
> ```
> 
> We can also expose `.response` so folks can read out response headers even before the data is streamed, for example

### LINKED ISSUE #778: 💡💭 make it simple to get custom response format for endpoint methods
Follow up for https://github.com/octokit/rest.js/issues/742#issuecomment-366563366

Endpoint methods such as `octokit.repos.getArchive({owner: 'octokit', repo: 'rest.js'})` currently resolve with JSON, text or a Buffer (ArrayBuffer in Browser), depending on the response headers. In some cases I might want to have more control of the response, e.g. I might want to get a readable stream or a buffer even for `application/json` responses.

I’m not yet sure what the best API would be that we can provide, but it would be nice to expose the request defaults that a method such as octokit.repos.getArchive sets, amend it with the client defaults such as baseUrl, default headers etc and then receive an object with request options that I can pass to my own request method, such as `fetch` in the browser or `http.request` in Node.

For example we could make something like this work

```js
const {url, ...options} = octokit.endpoint(
  octokit.repos.getArchive.defaults,
  {owner: 'octokit', repo: 'rest.js'}
)
const response = await fetch(url, options)
```

Or 

```js
const {url, ...options} = octokit.repos.getArchive.defaults({owner: 'octokit', repo: 'rest.js'})
const response = await fetch(url, options)
```

### LINKED ISSUE #824: trim payload
This is a feature request for doing less. 🙀 

I'm building an app using AWS Lambda and in order to keep coldstarts sub second we have a hard limit of 5mb uncompressed payloads. Unfortunately this module is over that size limit and I thought you'd like to know. Good news is I think if you can kick the lodash requirement, or at least be selective about which bits you use, the payload could be trimmed quite a bit. 

This is a scan of the package installed with `--production`.

screen shot 2018-03-23 at 2 05 07 pm

### LINKED ISSUE #829: Add method to register custom endpoint methods
`octokit` only has methods for the publicly documented APIs. But there are pre-releases that folks would love to try out, and we would love to make it as simple as possible for them. And if new features got released and they didn’t land in the latest `@octokit/rest` yet or someone would like to port it back to a previous version, they could do that.

So the idea is that GitHub can share JSON files for the pre-releases that follow the same format as [lib/routes.json](https://github.com/octokit/rest.js/blob/master/lib/routes.json).

Suggest for API 

```js
octokit.registerEndpoints({
  projects: {
    getAllCollaborators: {
      method: 'GET',
      path: '/projects/:id/collaborators',
      params: {
        "id": {
          "type": "string",
          "required": true,
        },
        "affiliation": {
          "type": "enum",
          "default": "all",
          "options": [
            "outside",
            "direct",
            "all"
          ]
        },
        "per_page": {
          "type": "number",
          "default": 30
        },
        "page": {
          "type": "number",
          "default": 1
        }
      }
    }
  }
})
```

This would be an endpoint for new project permission feature released yesterday: https://blog.github.com/2018-03-28-improved-organization-project-permissions/

Any thoughts on the API?

### LINKED ISSUE #988: get all references seems to be missing
I cannot find the endpoint for [get-all-references](https://developer.github.com/v3/git/refs/#get-all-references) either in the docs or in 15.10.0 . Please let me know where I can find it or if I can contribute a fix

### LINKED ISSUE #1068: Type for comment_id in TypeScript doesn't match documentation
I've been developing a GitHub-related Atom pluggin using TypeScript, and I've noticed a minor discrepency between the typings and the documentation. In the typings, the `comment_id` field (at least in issue functions) is a string, but in the documentation, the field is an integer. Is there any way that the typings could be changed to match the documentation or vice-versa.

### LINKED ISSUE #1069: 📯 Feedback wanted: Hooks API
`@octokit/rest` has already an experimental API to define hooks: https://github.com/octokit/rest.js/tree/33aba5d44d9a97a11b8077e443ec1c832bf8e15e/lib/plugins#usage

Before making the API public I’d love your input if it covers all use cases?

And a few questions:

- Can you think of a case when you’d need to remove a hook?
- Can you think of a case when you would need to define the hooks after the client is initiated?
- Any other use cases you’d like to implement with Hooks that are not shown below?
- Can you think of another "hook" than `'request'`? Can you think of plugins wanting to define their own hooks?

## Use cases

The code examples are simplified

1. [Authentication](#authentication)
1. [Validation](#validation)
1. [Retry](#retry)
1. [Throttling](#throttling)
1. [Caching](#caching)
1. [Mocking (testing)](#mocking)
1. [Use custom request logic](#custom-request-logic)
1. [Looging](#logging)

<a name="authentication"></a>
### Authentication

```js
client.hook.before('request', (options) => {
  // differentiate between basic, token, oauth
  // renew tokens for GitHub App installation if needed
})
```

<a name="validation"></a>
### Validation

```js
client.hook.before('request', (options) => {
  const params = getParamsForRoute(options)
  // validate passed options based on params
})
```

<a name="retry"></a>
### Retry
Retry all requests

```js
const pRetry = require('p-retry')
client.hook.wrap('request', (request, options) => {
  return pRetry(request.bind(null, options), {retries: 5})
})
```

Retry only for specific errors

```js
client.hook.wrap('request', async (request, options) => {
  try {
    return request(options)
  catch (error) {
    options.request.retries = (options.request.retries || 0) + 1
    if (options.request.retries > 5) {
      throw error
    }

    // retry on request errors
    if (isRequestError(error)) {
      return request(options)
    }

    throw error
  }
})
```

<a name="throttling"></a>
### Throttling

see [GitHub API Best Practises](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits)

```js
const Bottleneck = require('bottleneck')
const limiter = new Bottleneck({
  // Make requests for a single user or client ID serially.
  // Do not make requests for a single user or client ID concurrently.
  maxConcurrent: 1,

  // If you're making a large number of POST, PATCH, PUT, or DELETE requests
  // for a single user or client ID, wait at least one second between each request.
  minTime: 1000,

  // 5000 requests / hour limit
  reservoir: 5000, // max 5000 request per hour
  reservoirRefreshAmount: 5000,
  reservoirRefreshInterval: 60 * 60 * 1000 // every hour
})

client.hook.wrap('request', async (request, options) => {
  return limiter.schedule(request.bind(null, options))
})
```

<a name="caching"></a>
### Caching
Conditional requests

```js
const cache = {}
client.hook.after('request', (response, options) => {
  const eTag = response.headers['e-tag']
  if (eTag) {
    cache[response.headers['e-tag']] = response
  }
})
client.hook.after('error', (error, options) => {
  if (error.statusCode === 304) {
    return cache[options.headers['e-tag']]
  }

  throw error
})
```

<a name="mocking"></a>
### Mocking (testing)

```js
test('root request', (t) => {
  t.plan(2)
  client.hook.wrap('request', (request, options) => {
    t.equal(options.method, 'GET')
    t.equal(options.url, '/repos/octokit/rest.js')
  })
  return client.repos.get({owner: 'octokit', repo: 'rest.js'})
})
```

<a name="custom-request-logic"></a>
### Use custom request logic

```js
const axios = require('axios')
client.hook.wrap('request', (request, options) => {
  const requestOptions = client.request.endpoint(options)
  return axios(requestOptions)
})
```

<a name="logging"></a>
### Logging

```js
client.hook.wrap('request', (request, options) => {
  try {
    const result = await request(options)
    console.log('%s %s %d', options.method, options.url, response.statusCode)
  } catch (error_) {
    console.log('%s %s %d', options.method, options.url, error.response.statusCode)
  }
})
```

## Resources

- semantic-release reduced writing requests throttling from 1 per second to 1 per 3 seconds, see https://github.com/semantic-release/semantic-release/issues/843#issuecomment-405002364
- X-Poll-Interval: https://github.com/octokit/rest.js/issues/526#issuecomment-354179760
- semantic-release – implements throttling and retries
https://github.com/semantic-release/github/blob/d51858eba39b887a068e0d3fd1a61b107f2d04a6/lib/get-client.js

### LINKED ISSUE #1074: 📯 Feedback wanted: plugin API
The current `@octokit/rest` already has a plugin API, it’s just not documented yet. But plenty of folks use it already and so far I heard no complaints: https://github.com/octokit/rest.js/tree/33aba5d44d9a97a11b8077e443ec1c832bf8e15e/lib/plugins#readme

It’s as simple as it gets

```js
const octokit = require('@octokit/rest')()
octokit.plugin((octokit) => {
  octokit.newMethod = () => { octokit.log('hello world') }
})
octokit.newMethod() 
// logs "hello world"
```

Together with the `octokit.hook` API (see discussion at #1069) plugins will be able to extend Octokit’s functionality and people will be able to share plugins for common use cases, such as "edit multiple files in a pull request", etc.

I’m not yet sure if we need the `.plugin()` method on the instance, or if it would suffice to have it on the constructor, so we could do things like this

```js
const MyOctokit = require('@octokit/rest')
  .plugin(require('octokit-edit-pull-request-files'))
  .plugin(require('octokit-errors-sentry'), config)
```

Questions

- Can you think of use cases where you would need to add a plugin on runtime?
- Any feedback on the plugin AIP?

Thanks 🙏

gr2m:
Tracking v16 release

### Work in Progress

- [x] typescript definitions for static methods (not generated). Need to be added to [`/scripts/templates/index.d.ts.tpl`](https://github.com/octokit/rest.js/blob/223a38782abf5522d4c4c01c3b3da0429c66dddd/scripts/templates/index.d.ts.tpl)
  - [x] [octokit.request, octokit.request.endpoint, octokit.request.endpoint.merge, octokit.request.endpoint.parse](https://github.com/octokit/rest.js/blob/next/README.md#custom-requests)
  - [x] [octokit.hook, octokit.hook.before, octokit.hook.after, octokit.hook.error, octokit.hook.wrap](https://github.com/octokit/rest.js/blob/next/README.md#hooks)
  - [x] [octokit.plugin](https://github.com/octokit/rest.js/blob/next/README.md#plugins)
  - [x] [octokit.registerEndpoints](https://github.com/octokit/rest.js/blob/next/README.md#register-custom-endpoint-methods)

After release

- [ ] update `@octokit/rest` version in https://github.com/gr2m/octokit-pagination-methods/blob/master/package.json
- [ ] In release notes
  - Mention breaking change: `error.status` is now the response status code. The status text is no longer returned.
  - Mention to install with `^15.18.0` (`npm install @octokit/rest@15.18.0`) first, then run tests with ` | grep DEPRECATED` and rename all the methods / parameters, then install v16 and it should just work :tm: 
  - reduction of node_modules folder size (5mb+ to <1mb)
  - reduction of bundle size (~28kb to ~18kb)

### Done

- [x] #1094 Rename method names to align better with docs on https://developer.github.com/v3/. Deprecate old method names
- [x] line-by-line code review to remove all obsolete code
- [x] go through open issues & PRs
- [x] #1097 generate routes for real
- [x] remove tests for deprecations
- [x] https://github.com/octokit/request.js/pull/11 v16: `error.code` -> `error.status`, deprecate `error.code`.
- [x] #1091 deprecate callback APIs (remove from API docs)
- [x] #1092 deprecate `input` parameter in favour of `data`
- [x] #1093 deprecate `error.toJSON()`
- [x] #910 `.paginate()`
- [x] #910 deprecate v15 pagination methods
- [x] create `octokit-pagination-methods` package
- [x] move SCIM routes into plugin / separate repository. This will be a good showcase on how to use a build step to turn the routes format https://github.com/octokit/routes into the format that `octokit.registerEndpoints()` wants, while using the format from `ocotkit/routes` to generate documentation  (Repo: https://github.com/octokit/plugin-scim.js)
- [x] #829 Add method to register custom endpoint methods (Pull Request: #1086)
- [x] add all non-breaking changes from https://github.com/octokit/rest.js/pull/1043 to v15 (PR: https://github.com/octokit/rest.js/pull/1081)
- [x] make plugin API public (see discussion at #1074)
- [x] #1065 Use `@octokit/request` / `@octokit/endpoint` libraries in `@octokit/rest`. Make `client.request` and `client.request.endpoint` public APIs
- [x] implement new hooks API – see discussion at #1069
- [x] publish `@octokit/endpoint` (https://github.com/octokit/endpoint.js)
- [x] publish `@octokit/request` (https://github.com/octokit/request.js)
- [x] #684 parse JSON from GitHub Errors
- [x] #889 Rename routes to align better with docs on https://developer.github.com/v3/, PR: #1043 

### FAQ

- [Will it be possible to use a custom request library](https://github.com/octokit/rest.js/pull/1052#issuecomment-432623920)?

  Yes it will be possible, using the a new hooks API:

  ```js
  const axios = require('axios');
  github.hook.wrap('request', (request, options) => {
    return axios(options)
  })
  ```

---

closes #1068, closes #1069, closes #1074, closes #988, closes #829, closes #778, closes #777, closes #688, closes #687, closes #601, closes #766, closes #824


juliethefox:
Hello! May I ask what is meant by

> Use above libraries in `@octokit/rest` make `client.request` and `client.request.endpoint` public APIs

? 
With this release, will it be possible to use our custom http-client package (that has a slightly different interface from `node-fetch`)?

My question based on this [pull-request](https://github.com/octokit/rest.js/pull/1031)

gr2m:
> Hello! May I ask what is meant by

> > Use above libraries in `@octokit/rest` make `client.request` and `client.request.endpoint` public APIs

It’s just a refactoring, there will be no change in behavior. These are already implemented today, just part of the code base of `@octokit/rest`, see [lib/endpoint](https://github.com/octokit/rest.js/tree/44e9b1fb387c999e541e4f09d6face0fede33f45/lib/endpoint) and [lib/request)](https://github.com/octokit/rest.js/tree/44e9b1fb387c999e541e4f09d6face0fede33f45/lib/request).

> With this release, will it be possible to use our custom http-client package (that has a slightly different interface from node-fetch)?

Yes it will be possible, using the a new hooks API. You will be able to do something like this

```js
const got = require('got');
github.hooks((request, options) => {
  return got[options.method](options.url, options)
})
```

`request` in the above method is the original request method, but you can replace the request logic entirely as shown above. Would that work for you?

PR #1054, opened 2018-10-17
Update routes

gr2m:
closes #1053


gr2m:
This is blocked by the generated typescript Definition errors: https://travis-ci.org/octokit/rest.js/jobs/442450033#L568

I didn’t edit the script, I’m not sure what might have caused that. Could you have a look @gimenete?

gimenete:
@gr2m it looks like there are duplicated params with the same name in the `routes-for-api-docs.json` file.

screen shot 2018-10-17 at 16 28 43

For example there are two `data` params there ☝️ 

And that produces duplicated fields in the generated TypeScript

gr2m:
🤦‍♂️ apologies, I should have caught that myself, I’m looking int it

gimenete:
@gr2m no worries!

PR #1056, opened 2018-10-17
Fix for non-browser and non-node environments like React Native

### LINKED ISSUE #273: Add support for React Native
I attempted to use this package in an application I started building using React Native, unfortunately it did not work.

Requiring the package causes the react packager to hang.

```
var GitHubApi = require("github");
```

Pressing Command-R in the simulator produces (after about a minute):

```
Could not connect to the development server. Ensure node server is running and available on the same network - run 'nom start' from react-native root.

URL: http://localhost:8081/index.ios.bundle

The request timed out.
```

The packager hangs forever when attempting to debug.

I am told that not all packages run in React Native, mainly ones that designed specifically for Node JS (express for example).

brunolemos:
Fix #273

### DIFF
--- a/lib/parse-client-options.js
+++ b/lib/parse-client-options.js
@@ -67,7 +67,7 @@ function parseOptions (userOptions) {
   }
   /* istanbul ignore else */
 
-  if (!process.browser) {
+  if (!process.browser && typeof getRequestAgent === 'function') {
     clientDefaults.request.agent = getRequestAgent(clientDefaults.baseUrl, userOptions)
   }
 
--- a/lib/request/request.js
+++ b/lib/request/request.js
@@ -1,8 +1,11 @@
+/* global fetch */
+
 'use strict'
 
 module.exports = request
 
-const fetch = require('node-fetch').default
+const _fetch = typeof fetch === 'undefined' ? require('node-fetch').default : fetch
+
 const debug = require('debug')('octokit:rest')
 const defaults = require('lodash/defaults')
 const isPlainObject = require('lodash/isPlainObject')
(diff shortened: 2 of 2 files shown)


gr2m (commented):
Thanks so much for opening this PR! Getting it to work with react native would be great! 

Is react native’s "fetch" fully compatible with one in modern browsers? Can you think of a test we could add to the repository to make sure we don’t break react-native compatibility easily in future? Do you know a JS library that supports all 3 environments so I can see how they detect the different environments? I’m not sure about `process.title === 'node'`, is there something like `process.reactNative`?

brunolemos:
Changed the implementation, please re-check.

> Is react native’s "fetch" fully compatible with one in modern browsers? Can you think of a test we could add to the repository to make sure we don’t break react-native compatibility easily in future? Do you know a JS library that supports all 3 environments so I can see how they detect the different environments?

I don't know.

> not sure about process.title === 'node'

Changed to another way that also fixed the issue on react-native

> is there something like process.reactNative

I don't think we should make the solution too specific like that

> did you check if node-fetch could add compatibility with react native

They do have a [PR opened for 2 months](https://github.com/bitinn/node-fetch/pull/510) and I just made [another one](https://github.com/bitinn/node-fetch/pull/537) but they didn't show much interest in support different enviroments.

gr2m:
the upcoming version is now using https://github.com/octokit/request.js, could you please resubmit the pull request there? I’m okay with merging it for react native support until node-fetch lands it or it becomes obsolete.

I’d still love a test to make sure it actually works with react native. Maybe we can add a minimal react native app to the test directory and run tests on Travis, if that’s at all possible?

Thanks!

PR #1065, opened 2018-10-24
octokit.request & octokit.request.endpoint

gr2m:
part of https://github.com/octokit/rest.js/pull/1052

This requires breaking changes

- callbacks are no longer supported
- `input` parameter is now `data`
- `response.meta` (deprecated) is no longer returned
- error.toJSON() has been removed. error.toString() is no longer overwritten, the returned string now starts with `HttpError: `.


gr2m:
ugh this is failing by some very weird error that I cannot reproduce locally yet :(

image

PR #1083, opened 2018-11-05
Add mercy-preview header to apps.getInstallationRepositories

sjparkinson:
Enables the `topics` properties to be returned in the response of calls to `apps.getInstallationRepositories`.

See https://developer.github.com/v3/apps/installations/#installations for more information.

Resolves https://github.com/probot/probot/issues/760, though it's a round-about way.

### DIFF
--- a/lib/routes.json
+++ b/lib/routes.json
@@ -692,7 +692,7 @@
     },
     "getInstallationRepositories": {
       "headers": {
-        "accept": "application/vnd.github.machine-man-preview+json"
+        "accept": "application/vnd.github.mercy-preview+json,application/vnd.github.machine-man-preview+json"
       },
       "method": "GET",
       "params": {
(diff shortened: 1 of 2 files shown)


gr2m:
I only want to set the preview headers required for the APIs to work at all. You can set optional headers as you show [in your probot bug report](https://github.com/probot/probot/issues/760)

```js
octokit.apps.getInstallationRepositories({
  per_page: 100,
  headers: {
    accept: 'application/vnd.github.machine-man-preview+json,application/vnd.github.mercy-preview+json'
  }
})
```

It looks to me like Probot’s paginate method does not carry the custom headers to subsequent requests?

sjparkinson:
> It looks to me like Probot’s paginate method does not carry the custom headers to subsequent requests?

Yeah, seems to be the case, haven't figured it out 100%.

> I only want to set the preview headers required for the APIs to work at all.

Totally fair, thought that might be the case, but raised this just-in-case.

Thanks for having a look regardless.

PR #1092, opened 2018-11-09
deprecate options.input

gr2m:
Nobody should be affected by this, it’s an internal refactoring. But just in case, why not. Part of #1052.


gr2m:
merged via https://github.com/octokit/rest.js/commit/f7a2bae00b2c586456cebb29a46ba11216f74270

PR #1093, opened 2018-11-10
deprecate error.toJSON

gr2m:
part of #1052. You probably didn’t use that anyway :)


gr2m:
merged via https://github.com/octokit/rest.js/commit/a2844b618234b4b0d9e4021591a54f903e384f53

PR #1115, opened 2018-11-16
Adds date field to response headers

billytrend:
It would be nice to have this type available from typescript. I assume this is the correct way to add it?

Date appears to be in all github responses ie https://developer.github.com/v3/#schema

Could add some more fields if required


billytrend:
yeah of course, merge it however you like, thanks! BTW i was also going to make link optional since it only exists on paginated responses but this isn't high priority

PR #1117, opened 2018-11-18
Avoid Implicit Anys

### LINKED ISSUE #1116: Generated typings include implicit any
If the typescript compiler has the flag `noImplicitAny` enabled, some errors appear. These errors were introduced here https://github.com/octokit/rest.js/pull/1113

```
> tsc -p ./

node_modules/@octokit/rest/index.d.ts(22759,5): error TS7010: 'before', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@octokit/rest/index.d.ts(22760,5): error TS7010: 'after', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@octokit/rest/index.d.ts(22767,5): error TS7010: 'error', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@octokit/rest/index.d.ts(22771,5): error TS7010: 'wrap', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@octokit/rest/index.d.ts(22780,3): error TS7010: 'plugin', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@octokit/rest/index.d.ts(22782,3): error TS7010: 'registerEndpoints', which lacks return-type annotation, implicitly has an 'any' return type.

```

As far as I could see, all the hooks and `registerEndpoints` are `void`, and `plugin` returns a `Github` instance.

mamodom:
Closes #1116

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -201,15 +201,15 @@ declare class Github {
   authenticate(auth: Github.Auth): void;
 
   hook: {
-    before(name: string, callback: (options: Github.HookOptions) => void)
-    after(name: string, callback: (response: Github.Response<any>, options: Github.HookOptions) => void)
-    error(name: string, callback: (error: Github.HookError, options: Github.HookOptions) => void)
-    wrap(name: string, callback: (request: (options: Github.HookOptions) => Promise<Github.Response<any>>, options: Github.HookOptions) => void)
+    before(name: string, callback: (options: Github.HookOptions) => void): void
+    after(name: string, callback: (response: Github.Response<any>, options: Github.HookOptions) => void): void
+    error(name: string, callback: (error: Github.HookError, options: Github.HookOptions) => void): void
+    wrap(name: string, callback: (request: (options: Github.HookOptions) => Promise<Github.Response<any>>, options: Github.HookOptions) => void): void
   }
 
-  plugin(plugin: Github.Plugin | [Github.Plugin])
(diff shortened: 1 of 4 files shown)


scripts/templates/index.d.ts.tpl
```diff
@@ -201,15 +201,15 @@ declare class Github {
...
-    error(name: string, callback: (error: Github.HookError, options: Github.HookOptions) => void)
-    wrap(name: string, callback: (request: (options: Github.HookOptions) => Promise<Github.Response<any>>, options: Github.HookOptions) => void)
+    before(name: string, callback: (options: Github.HookOptions) => void): void
+    after(name: string, callback: (response: Github.Response<any>, options: Github.HookOptions) => void): void
+    error(name: string, callback: (error: Github.HookError, options: Github.HookOptions) => void): void
+    wrap(name: string, callback: (request: (options: Github.HookOptions) => Promise<Github.Response<any>>, options: Github.HookOptions) => void): void
   }
 
-  plugin(plugin: Github.Plugin | [Github.Plugin])
+  plugin(plugin: Github.Plugin | [Github.Plugin]): Github
```
gr2m: I think that was an oversight, there is no `.plugin` function on the instances, only on the class.

So you can do 

```js
const MyOctokit = require('@octokit/rest').plugin(require('my-octokit-plugin'))
```

but not

```js
const octokit = require('@octokit/rest')()
const octokit.plugin(require('my-octokit-plugin'))
```

/cc @gimenete

PR #1119, opened 2018-11-18
Include a header object param when transforming upload-release-asset's 'Content-Type' and 'Content-Length'

### LINKED ISSUE #1118: Missing `headers` property in `ReposUploadReleaseAssetParams` type definition
There is an error with [`upload-release-asset`](https://github.com/octokit/routes/blob/v15.0.1/routes/repos/releases/upload-a-release-asset.json)'s typings. I didn't find any way of linking to a version of `index.d.ts` so I'll include snippets here.

After installing [`@octokit/rest` v16.0.0](https://www.npmjs.com/package/@octokit/rest/v/16.0.0) and inspecting its `index.d.ts` there can be found at line 21678 this type declaration:
```
export type ReposUploadReleaseAssetParams = {
    url: string;
    /**
     * The file name of the asset. This should be set in a URI query parameter.
     */
    name: string;
    /**
     * An alternate short description of the asset. Used in place of the filename. This should be set in a URI query parameter.
     */
    label?: string;

    file: string | object;
  };
```

There are 2 params missing: [`Content-Type` and `Content-Length`](https://github.com/octokit/routes/blob/v15.0.1/routes/repos/releases/upload-a-release-asset.json#L15-L28). After some digging I saw that the tests included a `headers` object with the [2 missing parameters](https://github.com/octokit/rest.js/blob/v16.0.0/test/integration/release-assets-buffer-test.js#L32-L35)

I later found there is an [exception for `uploadReleaseAsset`](https://github.com/octokit/rest.js/blob/v16.0.0/scripts/lib/get-routes.js#L42-L59) that redefines `Content-Type` and `Content-Length` as `headers.content-type` and `headers.content-length`. 

After that figured out why `headers.content-type` and `headers.content-length` were being ignored when generating typings: 
If there isn't a object parameter `headers` (like [`create-hook`'s `config`](https://github.com/octokit/routes/blob/v15.0.1/routes/repos/hooks/create-a-hook.json#L29-L35)), they'll be treated as childParameters and be ignored when generating types.

My immediate reaction was to see how other header params were being handled elsewhere, but then I realized that [these 2 are the only ones that have `location: 'headers'`](https://runkit.com/embed/j7javgd5dv2g).

I'll submit 2 possible solitions.

mamodom:
Closes #1118 

~~Ignore this pull request if #1120 is merged~~ closes #1120

### DIFF
--- a/scripts/lib/get-routes.js
+++ b/scripts/lib/get-routes.js
@@ -54,7 +54,14 @@ function getRoutes () {
         indexes.forEach(i => endpoint.params.splice(i, 1))
         endpoint.params.unshift(
           Object.assign(contentLengthParam, { name: 'headers.content-length' }),
-          Object.assign(contentTypeParam, { name: 'headers.content-type' })
+          Object.assign(contentTypeParam, { name: 'headers.content-type' }),
+          {
+            name: 'headers',
+            location: 'headers',
+            required: true,
+            type: 'object',
+            description: 'Request headers containing `content-type` and `content-length`'
+          }
         )
       }
 


gr2m:
I’ll merge it locally to combine with #1117

gr2m (approved):
thanks!

PR #1120, opened 2018-11-18
Generically handle route params with location: 'headers'

### LINKED ISSUE #1118: Missing `headers` property in `ReposUploadReleaseAssetParams` type definition

mamodom:
closes #1118 

Ignore this pull request if #1119 is merged

### DIFF
--- a/scripts/lib/get-routes.js
+++ b/scripts/lib/get-routes.js
@@ -39,7 +39,6 @@ function getRoutes () {
 
     // handle some exceptions. TODO: move this into @octokit/routes
     endpoints.forEach(endpoint => {
-
       const headerParams = endpoint.params.filter(param => param.location === 'headers')
       if (headerParams.length) {
         headerParams.forEach(param => Object.assign(param, { name: `headers.${param.name.toLowerCase()}` }))
--- a/scripts/lib/get-routes.js
+++ b/scripts/lib/get-routes.js
@@ -39,23 +39,18 @@ function getRoutes () {
 
     // handle some exceptions. TODO: move this into @octokit/routes
     endpoints.forEach(endpoint => {
-      // exception for uploadReleaseAssets which passes parameters as header values
-      // see https://github.com/octokit/rest.js/pull/1043
-      if (endpoint.idName === 'uploadReleaseAsset') {
-        const contentLengthParam = endpoint.params.find(param => param.name === 'Content-Length')
-        const contentTypeParam = endpoint.params.find(param => param.name === 'Content-Type')
-        const indexes = endpoint.params.reduce((result, param, i) => {
-          if (['Content-Length', 'Content-Type'].includes(param.name)) {
-            result.unshift(i)
-          }
 
-          return result
-        }, [])
-        indexes.forEach(i => endpoint.params.splice(i, 1))
-        endpoint.params.unshift(
-          Object.assign(contentLengthParam, { name: 'headers.content-length' }),
-          Object.assign(contentTypeParam, { name: 'headers.content-type' })
-        )
+      const headerParams = endpoint.params.filter(param => param.location === 'headers')
+      if (headerParams.length) {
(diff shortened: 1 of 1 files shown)


scripts/lib/get-routes.js
```diff
@@ -39,23 +39,17 @@ function getRoutes () {
...
-          if (['Content-Length', 'Content-Type'].includes(param.name)) {
-            result.unshift(i)
-          }
+      const headerParams = endpoint.params.filter(param => param.location === 'headers')
+      if (headerParams.length) {
+        headerParams.forEach(param => Object.assign(param, { name: `headers.${param.name.toLowerCase()}` }))
 
-          return result
-        }, [])
-        indexes.forEach(i => endpoint.params.splice(i, 1))
```
gr2m, line 54: I used the `indexes` variable so that the parameters remain in the same place. That is relevant for the docs that are also generated from this script.

scripts/lib/get-routes.js
```diff
@@ -39,23 +39,17 @@ function getRoutes () {
 
     // handle some exceptions. TODO: move this into @octokit/routes
     endpoints.forEach(endpoint => {
-      // exception for uploadReleaseAssets which passes parameters as header values
-      // see https://github.com/octokit/rest.js/pull/1043
```
gr2m, line 43: I like your more generic script, but could you please keep a variation of the comment above to explain why the code is here? You could mention `uploadReleaseAssets` as en example

gr2m (commented):
This is great work, thank you Maximo! I just have two remarks, let me know what you think!

scripts/lib/get-routes.js
gr2m: Ah 🤦‍♂️ now I know why you removed that. Okay in that case I think I’d go with #1119. If there will be more than one endpoints in future we can  still make the implementation more generic. It’s a problem when it’s a problem :)

PR #1134, opened 2018-11-25
Update routes.json

hipstersmoothie:
In v16.0.0 it the types said that pullRequest didn't exists and pulls does. When running the code this wasn't the case. Changing this fixed it.


gr2m:
Oh my sorry for that, I'll have a look into this tomorrow!

PR #1136, opened 2018-11-26
fix(package): set `node-fetch` as a `devDependencies`

pvdlg:
This moves the `node-fetch` to `devDependency` as it is used in the test but not in the production code.
That solves the `HttpError: response.headers.entries(...) is not a function or its return value is not iterable`.

With the current setting, npm produces the following tree (in case something else in the project depends on `node-fetch@^1.0.0`):
```
my-project
├─┬ @octokit/rest@16.0.1
│ └── node-fetch@2.3.0
└── @octokit/request@2.1.12
└── node-fetch@1.7.3
```

So because `node-fetch@2.3.0` is depended upon in `@octokit/rest` it is not added in `@octokit/request/node_modules`. So when `@octokit/request` require `node-fetch` it's resolved by going up the tree and end up using the one at the root of the project which might be an older version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,43 +64,6 @@
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
         "js-tokens": "^3.0.0"
-      },
-      "dependencies": {
-        "ansi-styles": {
-          "version": "3.2.1",
-          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
-          "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
-          "dev": true,
-          "requires": {
-            "color-convert": "^1.9.0"
-          }
-        },
-        "chalk": {
-          "version": "2.4.1",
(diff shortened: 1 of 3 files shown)


gr2m:
Thanks @pvdlg! I think I can remove the node-fetch dependency altogether

pvdlg:
It's used there: https://github.com/octokit/rest.js/blob/6713e41a5c73ad3d678033b9e54b0ab4d6a204b3/test/util.js#L7

gr2m:
I got it

pvdlg:
Your commit https://github.com/octokit/rest.js/pull/1136/commits/9518e25ca86dc7c983ae1bb77a3c0fc01c7c66fa somehow changed a few `https` URLs into `http` in `package-lock.json`. Don't know or why but seems weird.

gr2m:
that’s very curious indeed! I re-created it but it still does that. I’ll go ahead and merge this PR but will ping npm about it

pvdlg:
See https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818 and https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

gr2m:
I fixed it manually now but I’m not sure how big the risk is and if it’s worth to put in the effort to fix the `http://` links in postinstall. I’ll probably just follow https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/8. Thanks again :)

PR #1138, opened 2018-11-26
Fix authenticate function

### LINKED ISSUE #1078: Typescript for the authenticate() function wrong
In the outputted TypeScript, you use the custom type which is the union of all the authentication methods types.
In theory that's good, but in practice it isn't. I'm writing an application and use that function, and linters complain that the passed params in the options object are missing or invalid.

All in, all this should be what the definition for that function looks like for it to work properly:
```ts
authenticate(auth: Github.AuthBasic): void;
authenticate(auth: Github.AuthOAuthToken): void;
authenticate(auth: Github.AuthOAuthSecret): void;
authenticate(auth: Github.AuthUserToken): void;
authenticate(auth: Github.AuthJWT): void;
authenticate(auth: Github.DeprecatedAuthJWT): void;
```

wolfy1339:
Use Typescript overloads instead of a union of different interfaces
Fixes #1078

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -105,13 +105,6 @@ declare namespace Github {
     token: string;
   }
 
-  export type Auth =
-    | AuthBasic
-    | AuthOAuthToken
-    | AuthOAuthSecret
-    | AuthUserToken
-    | AuthJWT;
-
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thanks @wolfy1339!

PR #1142, opened 2018-11-27
Don't forget await!

fabiosantoscode:
-----
[View rendered README.md](https://github.com/fabiosantoscode/rest.js/blob/patch-1/README.md)


gr2m (approved):
Nice catch!

PR #1156, opened 2018-12-12
Update to @octokit/routes 16.0.0

gr2m:
closes #1155


gr2m:
There are no relevant changes

PR #1176, opened 2018-12-29
Paginate with link header pointing to different path

gr2m:
seen here: https://github.com/probot/probot/blob/bb06c51485245d75508982826180e7c2e774a7db/test/github.test.ts#L40-L56

closes #1178


gr2m:
This will pass once https://github.com/octokit/fixtures/pull/193 lands in octokit-fixtures-server

PR #1180, opened 2019-01-03
Update TypeScript definitions for API functions

wolfy1339:
~~This makes it easier to use the definitions so you can only see relevant information if you're using callbacks or Promise's.
Also allows for better understanding of the code by linters. Before, one could try to mix the 2 methods together and linters wouldn't catch it. Now they would~~
Removes the callback option from the TypeScript definitions as they are no longer supported, and marks the params as optional, since they aren't needed as per the docs


gr2m:
We no longer support callbacks for the endpoint methods, I guess we missed that in the Typescript definitions?

wolfy1339:
Alright then. I will fix that up

wolfy1339:
I have fixed it up. And while looking through the TravisCI logs, I've seen that your API token is exposed and is not secret. It is logged as plaintext

gr2m (approved):

PR #1186, opened 2019-01-08
Pagination: early exit

gr2m:
Besides the current `response`, the map function now also receives a `done()` function as 2nd argument. You can call it for an early exit. You still have to return the value you want to map the response to

```js
octokit.paginate('GET /organizations', (response, done) => {
  if (response.data.find(issues => issue.body.includes('something'))) {
    done()
  }
  return response.data
})
```

-----
[View rendered README.md](https://github.com/octokit/rest.js/blob/pagination-early-exit/README.md)


gr2m:
merged via 2057d2a

PR #1195, opened 2019-01-11
Update to @octokit/routes v18.1.1

gr2m:
Closes #1193

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,5 @@
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-    packages:
-      - g++-4.8
 language: node_js
-cache:
-  directories:
-    - ~/.npm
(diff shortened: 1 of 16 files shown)


gr2m:
I need to implement deprecations again :) Now the deprecations live in `@octokit/routes`. In this case 
`get-or-create-authorization-for-app-fingerprint` has been changed to `get-or-create-authorization-for-app-and-fingerprint`. So now there are two methods, the latter replacing the former. But we don’t want to just remove the former as it would require a breaking change. Instead we want to log a deprecation message to help people migrate to the new methods. 

Here is the definition in `@octokit/routes`: https://github.com/octokit/routes/blob/268049721ab5db75eb04a87ff5d41ff44aa4b0d1/routes/api.github.com/oauth-authorizations/get-or-create-authorization-for-app-fingerprint.json#L63-L72

PR #1197, opened 2019-01-12
fix: do not load the same plugin again

SGrondin:
At the moment, users of the throttling and retry plugins are installing and loading them manually. Once we include those plugins into the core library, it's going to cause issues. For example, having the retry plugin loaded twice would mean failed requests would be retried 6 times instead of 3.

This PR simply skips a plugin if it's already present. Should it also display a warning?

### DIFF
--- a/test/integration/plugins-test.js
+++ b/test/integration/plugins-test.js
@@ -11,7 +11,7 @@ describe('plugins', () => {
     expect(myClient.foo).to.equal('bar')
   })
 
-  it('it does not override plugins of original constructor', () => {
+  it('does not override plugins of original constructor', () => {
     const MyOctokit = Octokit.plugin((octokit) => {
       octokit.foo = 'bar'
     })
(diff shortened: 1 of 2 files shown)


test/integration/plugins-test.js
```diff
@@ -28,4 +28,16 @@ describe('plugins', () => {
...
+    const myPlugin = (octokit, options) => {
+      if (octokit.customKey) {
+        throw new Error('Boom!')
+      } else {
+        octokit.customKey = true
+      }
+    }
+    const MyOctokit = Octokit.plugin(myPlugin).plugin(myPlugin)
+    expect(MyOctokit).to.not.throw()
+  })
```
gr2m, line 42: nice test 👍

SGrondin:
Note: this isn't foolproof. If a user has a dependency on `^2.1.0` and octokit/rest ships with `^3.0.0` then the same plugin will be loaded twice. That's why I think it would be nice if plugins had to provide a name.

But overall this is much better than nothing!

gr2m (approved):
Thanks!

PR #1199, opened 2019-01-16
Add memory leakage test

gr2m:
After hunting a memory leak in the latest Octokit for a while (I turned out to be the throttle/reply plugin), I want to start running tests that detect potential memory leaks so none can sneak in in future. I’m open to ideas, running memory tests on CI is new to me!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,3 @@
-addons:
-  apt:
-    sources:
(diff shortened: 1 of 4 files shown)


gr2m:
done via c697a45acd08574e4d1738f5a3f84c34eb4f738b

PR #1204, opened 2019-01-17
two-factor authentication

### LINKED ISSUE #669: github basic authentication does not support 2FA
If a user has 2FA enabled, the authenticate() call will fail. This example will reproduce the issue:

1. Enable 2FA on your account
2. Run the basic auth example
3. Make any other call, for example request user info
```
github.authenticate({
  type: 'basic',
  username: process.env.USERNAME,
  password: process.env.PASSWORD
})

github.users.getForUser({username: 'jtrinklein'})
```

This will fail with the following error:

```
{
  message: '{"message":"Must specify two-factor authentication OTP code.","documentation_url":"https://developer.github.com/v3/auth#working-with-two-factor-authentication"}',
  code: 401,
  status: 'Unauthorized',
  headers:
  { server: 'GitHub.com',
     status: '401 Unauthorized',
     'x-github-otp': 'required; app',
     <snip>
  }
}
```

Github documentation for [working with two factor authentication](https://developer.github.com/v3/auth#working-with-two-factor-authentication) says:

> For users with two-factor authentication enabled, Basic Authentication requires an extra step. When you attempt to authenticate with Basic Authentication, the server will respond with a 401 and an X-GitHub-OTP: required; :2fa-type header. This indicates that a two-factor authentication code is needed (in addition to the username and password). The :2fa-type in this header indicates whether the account receives its two-factor authentication codes via SMS or via an application.
> 
> In addition to the Basic Authentication credentials, you must send the user's authentication code (i.e., one-time password) in the X-GitHub-OTP header.

✏️  Forgot to specify you need to make an actual API request for the authentication failure to manifest.

gr2m:
You can now pass a `on2fa` method to `octokit.authenticate` which gets called when a 2FA code is requested

```js
// basic + two-factor
octokit.authenticate({
  type: 'basic',
  username: 'yourusername',
  password: 'password'
  on2fa () {
    // must return or resolve with a two-factor code
  }
})
```

closes #669

-----
[View rendered README.md](https://github.com/octokit/rest.js/blob/two-factor-auth/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,16 @@ octokit.authenticate({
   password: 'password'
 })
 
+// basic + two-factor
+octokit.authenticate({
+  type: 'basic',
+  username: 'yourusername',
+  password: 'password'
+  on2fa () {
+    // must return or resolve with a two-factor code
+  }
+})
+
 // oauth
 octokit.authenticate({
   type: 'oauth',
--- a/plugins/authentication/index.js
+++ b/plugins/authentication/index.js
@@ -2,11 +2,14 @@ module.exports = authenticationPlugin
 
 const authenticate = require('./authenticate')
 const beforeRequest = require('./before-request')
+const requestError = require('./request-error')
 
 function authenticationPlugin (octokit) {
   const state = {
+    octokit,
     auth: false
   }
   octokit.authenticate = authenticate.bind(null, state)
   octokit.hook.before('request', beforeRequest.bind(null, state))
+  octokit.hook.error('request', requestError.bind(null, state))
 }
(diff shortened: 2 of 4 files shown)


mattfysh:
@gr2m - trying to use this snippet, and it keeps asking me for my 2FA code. Let me know if I'm doing something wrong.

```
octokit.authenticate({
  ...ghAuth,
  on2fa: () => new Promise(resolve => {
    rl.question('Enter your GitHub 2FA code: ', resolve)
  })
})
```

gr2m:
Hmm, does it end up in a loop? I thought I fixed that via https://github.com/octokit/rest.js/commit/d29ec5aca70c55a04cdccb020ac0cfaa3f256ebd 🤔 

I added a test to make sure that async on2fa works as expected: https://github.com/octokit/rest.js/pull/1213

I cannot reproduce the problem yet I’m afraid

PR #1211, opened 2019-01-21
useragent, previews and request options

gr2m:
In preparation for v17, instead of

```js
const octokit = new Octokit({
  headers: {
    accept: 'application/vnd.github.jean-grey-preview+json,application/vnd.github.symmetra-preview+json',
    'user-agent': 'octokit/rest.js v1.2.3' // v1.2.3 will be current version
  },
  timeout: 0,
  agent: undefined
})
```

the new options will be

```js
const octokit = new Octokit({
  userAgent: 'myapp v1.2.3',
  previews: [
    'jean-grey-preview',
    'symmetra-preview'
  ],
  request: {
    timeout: 0,
    agent: undefined
  }
})
```

The old options will continue to work, but will be deprecated. The `baseUrl` option will remain the same.

-----
[View rendered README.md](https://github.com/octokit/rest.js/blob/useragent-previews-options/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,14 @@
 Install with `npm install @octokit/rest`.
 
 ```js
-const octokit = require('@octokit/rest')()
+const Octokit = require('@octokit/rest')
+const octokit = new Octokit ()
 
 // Compare: https://developer.github.com/v3/repos/#list-organization-repositories
 octokit.repos.listForOrg({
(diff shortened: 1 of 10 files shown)


gr2m:
done via https://github.com/octokit/rest.js/releases/tag/v16.12.0

PR #1212, opened 2019-01-21
Update getForOrg to listForOrg in example

### LINKED ISSUE #1209: Uncaught TypeError: octokit.repos.getForOrg is not a function from README example
In the README file the example 

```
const octokit = new Octokit()

// Compare: https://developer.github.com/v3/repos/#list-organization-repositories
octokit.repos.getForOrg({
  org: 'octokit',
  type: 'public'
}).then(({data, headers, status}) => {
  // handle data
})
```

is given [here](https://github.com/octokit/rest.js#browser), however if I run that example I get the error 

> Uncaught TypeError: octokit.repos.getForOrg is not a function

When I inspect the ocotokit object it does not have a getForOrg method, and that method is also not listed in [the rest.js documentation](https://octokit.github.io/rest.js/), am I doing something wrong or is the README out of date?

BarbourSmith:
Update the browser example in the README to reflect the changes to the API method name

Fixes #1209

-----
[View rendered README.md](https://github.com/BarbourSmith/rest.js/blob/update-readme-example-topic/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ octokit.repos.listForOrg({
    const octokit = new Octokit()
 
    // Compare: https://developer.github.com/v3/repos/#list-organization-repositories
-   octokit.repos.getForOrg({
+   octokit.repos.listForOrg({
      org: 'octokit',
      type: 'public'
    }).then(({data, headers, status}) => {


gr2m (approved):
Perfect, thank you so much 💐

PR #1215, opened 2019-01-21
new Octokit({auth})

### LINKED ISSUE #1207: new Octokit({auth})
The current `octokit.authenticate` method is problematic in that it can change the type authentication within the same Octokit instance. I don’t see a valid use case where I would create an Octokit instance for a user and then later would change the authentication to a GitHub app. I would create separate Octokit instances instead.

To address that I issue, I would suggest to add an `auth` option to the constructor.

```js
const clientWithAuth = new Octokit({
  auth: 'token secret123'
})
```

The `auth` option can be

1. A string

   The value will be passed as value for the `Authorization` header,
   see [authentication](https://developer.github.com/v3/#authentication).

   ```js
   new Octokit({
     auth: 'token secret123'
   })
   ```

2. As object with the properties `username`, `password`, `on2fa`.

   `on2fa` is an asynchronous function that must resolve with two-factor
   authentication code sent to the user.

   ```js
   new Octokit({
     auth: {
       username: 'octocat',
       password: 'secret',
       async on2fa () {
         // example: ask the user
         return prompt('Two-factor authentication Code:')
       }
     }
   })
   ```

3. An object with the properties `client_id` and `client_secret`

   OAuth applications can authenticate using their `client_id` and `client_secret`
   in order to [increase the unauthenticated rate limit](https://developer.github.com/v3/#increasing-the-unauthenticated-rate-limit-for-oauth-applications).

4. A function

   Must resolve with a string which then will be passed as value for the
   `Authorization` header. The function will be called before each request and
   can be asynchronous.

   ```js
   new Octokit({
     auth () {
       return 'token secret123'
     }
   })
   ```

There is a special case for OAuth applications, when `client_id` and `client_secret` is passed as Basic Authorization instead of query parameters. The only routes where that applies share the same URL though: `/applications/:client_id/tokens/:access_token`. It should be easy enough to implement an exception within Octokit. The endpoints are

   1. [Check an authorization](https://developer.github.com/v3/oauth_authorizations/#check-an-authorization)
   1. [Reset an authorization](https://developer.github.com/v3/oauth_authorizations/#reset-an-authorization)
   1. [Revoke an authorization for an application](https://developer.github.com/v3/oauth_authorizations/#revoke-an-authorization-for-an-application)

For the record, I tested if these endpoints really require client_id & client_secret to be passed as Basic authorization or if they can also be passed as query arguments like recommended in the REST API docs for [Increasing the unauthenticated rate limit for OAuth applications](https://developer.github.com/v3/#increasing-the-unauthenticated-rate-limit-for-oauth-applications).

This will deprecate the `octokit.authenticate` method. Please let me know if I am missing any relevant use cases which are not well supported with this new API.

Related: #898

gr2m:
This deprecates `octokit.authenticate()`. Let me know if there is a use case that is not well covered by the new API.

closes #1207, addresses #998

-----
[View rendered README.md](https://github.com/octokit/rest.js/blob/1207/auth-option/README.md)


gr2m:
merged via 6ee188a

PR #1223, opened 2019-01-25
docs: add support for searching with api.github.com's url in gh-pages

soonoo:
Since there are so many endpoints in GitHub API, it's pretty annoying to me for finding correct endpoints.
When I type `repo` in [https://octokit.github.io/rest.js/](https://octokit.github.io/rest.js/) to get interface of `/user/repo`, plenty of items are displayed like below.  (about 30 ~ 40 items are in search result)  

  
  
  
I think it would be comfortable if searching with GitHub's API url (ex. `/user/repo`) is possible.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -275,6 +275,10 @@ pre code.sample-request-response-json {
   font-size: 14px;
 }
 
+.sidenav > li > span {
+  display: none;
+}
+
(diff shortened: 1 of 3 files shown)


gr2m (approved):
This is great, thank you!

PR #1224, opened 2019-01-26
Update docs template

soonoo:
rest.js uses `apidoc` module to create gh-pages.
`apidoc` module's [default templates](https://github.com/apidoc/apidoc/tree/master/template) are used when running `postbuild:docs` script.

Changes in #1223 are applied to default template of apidoc's template and imported in `docs_template` directory.


soonoo:
Test fails but I don't know how to exclude docs_template/ directory from being tested.
Can anyone explain about how to do this please?

gr2m:
We use [`standard`](https://standardjs.com) for linting. You can configure it to ignore files by adding a key to package.json: https://standardjs.com/#how-do-i-ignore-files

```json
"standard": {
  "ignore": [
    "docs_template/",
  ]
}
```

gr2m:
By the way, I’m working on a replacement for the docs using [GatsbyJS](https://www.gatsbyjs.org/): https://github.com/gr2m/gatsby-api-docs-playground in case you are interested :) I could definitely use some more help there

PR #1225, opened 2019-01-26
fix(typescript): accept `{ auth: () =>  }` constructor option

gr2m:
Instead of `{ auth: { auth: () =>  } }`


wolfy1339:
Ah, my bad

PR #1227, opened 2019-01-28
Fix README case format for clientId, clientSecret

cooperka:
Code usage in [authentication/validate.js](https://github.com/octokit/rest.js/blob/master@%7B2019-01-27%7D/plugins/authentication/validate.js#L16) and [authentication/before-request.js](https://github.com/octokit/rest.js/blob/master@%7B2019-01-27%7D/plugins/authentication/before-request.js#L25).

[View rendered README.md](https://github.com/cooperka/rest.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,9 +160,9 @@ The `auth` option can be
    })
    ```
 
-3. An object with the properties `client_id` and `client_secret`
+3. An object with the properties `clientId` and `clientSecret`
 
-   OAuth applications can authenticate using their `client_id` and `client_secret`
+   OAuth applications can authenticate using their `clientId` and `clientSecret`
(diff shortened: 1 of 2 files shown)


gr2m:
Oh my, good catch! Thanks for the pull request! Yes, Typescript definitions only need to updated at 
https://github.com/octokit/rest.js/blob/0e116993109603cfaa6b2d00feacf27a1e2dec5b/scripts/templates/index.d.ts.tpl#L45-L47

cooperka:
@gr2m done :hibiscus:

wolfy1339 (approved):
Thanks 👍

gr2m (approved):
Thanks 💐

gr2m:
Thanks for the review @wolfy1339 👍

Issue #1189, opened 2019-01-09
TS2304: Cannot find name 'AsyncIterableIterator'


ranyitz:
👋 Hey!

When running `tsc` I now get a new error. (starting from version `16.7.0`)

```
node_modules/@octokit/rest/index.d.ts(168,10): error TS2304: Cannot find name 'AsyncIterableIterator'.
```

In order to fix it, I needed to add `ESNext.AsyncIterable` to `tsconfig` `lib` option.

It's a breaking change and it's not clear from the docs that it should be added.

gr2m:
👋 Hey Ran, thanks for opening the issue!

I wouldn’t consider typescript changes as breaking changes because they only occur on build times, not in production.

The new Typescript additions describes the ability for async iterators when paginating

```js
for await (const response of octokit.paginate.iterator(options)) {
  // do whatever you want with each response, break out of the loop, etc.
}
```

You don’t have to use the feature, but if you do Typescript will no longer complain.

Is there anything I can add to make this error go away, or do you have to add the configuration on your side?

gr2m:
After researching this for a bit, I guess the answer to the error is to add `esnext.asynciterable` to `"lib": []` option in the `tsconfig.json` file

```json
{
  "lib": ["esnext.asynciterable"] 
}
```

PR #1229, opened 2019-01-29
Improve plug'n'playability of TS declarations

weswigham:
This declaration file uses the esnext async iterable types - if using [TS >= 3.0, this comment will pull that lib in automatically.](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#-reference-lib--reference-directives)

Related: #1189

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -6,7 +6,6 @@
 
 import * as http from 'http'
 
-
 declare namespace Octokit {
   type json = any
   type date = string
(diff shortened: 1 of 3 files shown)


weswigham:
Geh, older versions of TS issue an error on the directive - just keep this in mind if you update to a newer TS version.

gr2m:
Okay so updating typescript to latest version should take care of it? Can you add it to this PR?

scripts/templates/index.d.ts.tpl
```diff
@@ -2,6 +2,8 @@
  * This declaration file requires TypeScript 2.1 or above.
```
wolfy1339:
```suggestion
  * This declaration file requires TypeScript 3.0 or above.
```

wolfy1339 (changes requested):

gr2m (approved):
Thanks everyone!

PR #1230, opened 2019-01-30
Add listTopics preview header

leifdreizler:
This PR adds the required [preview header](https://developer.github.com/v3/repos/#list-all-topics-for-a-repository) to the [repos.listTopics](https://octokit.github.io/rest.js/#api-Repos-listTopics) function.

While using the `@octokit/rest` package, I noticed I was getting `415`s on the `repos.listTopics` request. After reviewing the docs, I changed my local copy of the `@octokit/rest` package to include the `mercy-preview` header, and the requests completed as expected.

### DIFF
--- a/plugins/rest-api-endpoints/routes.json
+++ b/plugins/rest-api-endpoints/routes.json
@@ -8376,6 +8376,9 @@
       "url": "/repos/:owner/:repo/teams"
     },
     "listTopics": {
+      "headers": {
+        "accept": "application/vnd.github.mercy-preview+json"
+      },
       "method": "GET",
       "params": {
         "owner": {


leifdreizler:
Thanks for the quick reply!

It's certainly possible that I'm messing something up.

But, without my change to `routes.json` I was getting a 415 with: 
`HttpError: If you would like to help us test the Repository Topics API during its preview period, you must specify a custommedia type in the 'Accept' header. Please see the docs for full details.`
and without it I was getting the repo topics returned.

I'm happy to reach out to support 👍🏻

leifdreizler:
I submitted this to GH support (#76473), received the following response:

A possible work around may be to shorten the string used to detect the presence of an Accept header to:

> you must provide a custom media type in the Accept header

gr2m:
I’m tracking the issue here: https://github.com/octokit/routes/issues/356. Once that is resolved, the `plugins/rest-api-endpoints/routes.json` file will get updated accordingly

PR #1231, opened 2019-01-30
Fix installation authentication example

pascal-giguere:
The README shows an incorrect example of authentication using a function.

`app.getInstallationAccessToken()` returns just the token, whereas `auth` expects the token to be preceded by `"token "`.

-----
[View rendered README.md](https://github.com/pascal-giguere/rest.js/blob/fix-installation-auth-doc/README.md)


gr2m (approved):
Great catch, thanks a lot!

PR #1240, opened 2019-02-12
If the auth method throws the exception gets lost

jasonk:
Since the `authenticate` method was deprecated, I updated my code to use an `auth` method to provide the token, like so:

```js
const octokit = new Octokit( {
  baseUrl,
  auth() { return `token ${this.getAuthToken()}`; },
} );
```

Unfortunately, this new version had a subtle bug (wrong `this` context) that took some time to track down because if the `auth` function throws an exception it gets hidden, because the error gets passed to `authenticationRequestError` which assumes that it will have a `headers` object, so any error thrown from the auth function gets turned into `TypeError: Cannot read property 'x-github-otp' of undefined`.

This one-line fix to just rethrow the error if it doesn't have headers seems to solve the problem for me, and it looks to me like this was the intended behavior of this function, so hopefully this helps.

### DIFF
--- a/plugins/authentication/request-error.js
+++ b/plugins/authentication/request-error.js
@@ -3,6 +3,7 @@ module.exports = authenticationRequestError
 const HttpError = require('@octokit/request/lib/http-error')
 
 function authenticationRequestError (state, error, options) {
+  if (!error.headers) throw error
   const otpRequired = /required/.test(error.headers['x-github-otp'] || '')
   // handle "2FA required" error only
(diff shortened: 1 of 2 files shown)


gr2m (commented):
Thank you so much for not only reporting the problem, but taking the time to understand the code and submitting a pull request.

Now the CI fails because we lost 100% code coverage. Could you add a test? It would go into https://github.com/octokit/rest.js/blob/master/test/integration/authentication-test.js. Let me know if you need any help with that

PR #1249, opened 2019-02-21
Update authentication examples

hudovisk:
The doc is pointing to the package `octokit`, which is not recommended yet.

Also updated the examples with proper format for the Authorization header.

-----
[View rendered README.md](https://github.com/hudovisk/rest.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ The `auth` option can be
 
    ```js
    const App = require('@octokit/app')
-   const Octokit = require('octokit')
+   const Octokit = require('@octokit/rest')
 
    const app = new App({ id: process.env.APP_ID, privateKey: process.env.PRIVATE_KEY })
    const octokit = new Octokit({


README.md
```diff
@@ -193,15 +193,15 @@ The `auth` option can be
 
    ```js
    const App = require('@octokit/app')
-   const Octokit = require('octokit')
+   const Octokit = require('@octokit/rest')
```
gr2m, line 196: Ooops, good catch!

README.md
```diff
@@ -127,7 +127,7 @@ To enable authenticated requests, pass an `auth` option to the Octokit construct
 
 ```js
 const clientWithAuth = new Octokit({
-  auth: 'token secret123'
+  auth: 'Bearer secret123'
```
gr2m: `Bearer` is used when authenticing as a GitHub Application, otherwise it’s token. But I think they might be interchangable, they at least are for the GraphQL API, I just tried that today. But `token` is definitely right here and more common in GitHub’s own API

README.md
```diff
@@ -193,15 +193,15 @@ The `auth` option can be
...
+   const Octokit = require('@octokit/rest')
 
    const app = new App({ id: process.env.APP_ID, privateKey: process.env.PRIVATE_KEY })
    const octokit = new Octokit({
      async auth () {
        const installationAccessToken = await app.getInstallationAccessToken({ 
          installationId: process.env.INSTALLATION_ID 
        });
-       return `token ${installationAccessToken}`;
+       return `Bearer ${installationAccessToken}`;
```
gr2m: I’m pretty sure that here it should definitely be token

hudovisk: You are rigth, but i guess it is also interchangable, I do have one app that uses `Bearer`. Anyway I updated to the correct form. Thanks

gr2m: `Bearer` is used for authentication as the GitHub app itself. `token` is used to authenticate as an app’s installation on a GitHub account or organization

gr2m (approved):
Thanks 👍

PR #1268, opened 2019-03-07
update @octokit/routes to 18.7.0

gr2m:
closes #1265


gr2m (commented):
feat: `octokit.licenses.listCommonlyUsed()`

`octokit.licenses.list()` is now deprecated

Issue #1279, opened 2019-03-13
on2fa prompts multiple times


ruxandrafed:
Hi guys,

I updated a CLI project I was working on to use the latest version, and I'm running into some odd issues re. `on2fa`.

Inside a separate function, I am creating the client like this:
```
const createAuthedClient = async (credentials) => {
  try {
    const client = new Octokit({
      auth: {
        username: credentials.username,
        password: credentials.password,
        async on2fa () {
          const { otp } = await inquirer.askGitHubOTP()
          return otp
        }
      }
    })
    return client
  } catch (error) {
    throw error
  }
}
```

and `inquirer.askGitHubOTP` only does the following:
```
const askGitHubOTP = () => {
  const questions = [
    {
      name: 'otp',
      type: 'input',
      message: 'Your GitHub OTP/2FA Code:',
      filter: input => input.trim()
    }
  ]
  return inquirer.prompt(questions)
}
```

I only run this once in my flow, but for some reason, the `on2fa` method ends up executing 4 times, so I get prompted 4 times for the code. As long as I type the right one, it keeps going .. I must say, also, that it occasionally happened to prompt me just once or twice.

What am I doing wrong? Any tips? Thank you!!

gr2m:
Hi Ruxandra 👋 

I don’t see any problem right away 🤔 Is the full code somewhere I could have a look at it? If not, could you set up a minimal repository with only the on2fa handling + inquirer?

Can you set logging as described at https://github.com/octokit/rest.js#debug, maybe it gives us some more information.

ruxandrafed:
Hey Gregor, thanks for the prompt response!

I set up logging and noticed that the `on2fa` actually seems to run when I use other authorization methods. Let me take a step back and explain what this CLI is trying to do:

1. Prompt user how they want to authenticate - token or user/pass
2. If user picks user/pass and has 2FA, get credentials and 2FA code
3. Create the client using the credentials.
4. Generate a token with the `repo` and `read:org` scopes (using [createAuthorization](https://octokit.github.io/rest.js/#api-OauthAuthorizations-createAuthorization))
5. If the token already exists, list all authorizations (using [listAuthorizations](https://octokit.github.io/rest.js/#api-OauthAuthorizations-listAuthorizations)) and find the auth id for it
6. Delete the old authorization (using [deleteAuthorization](https://octokit.github.io/rest.js/#api-OauthAuthorizations-deleteAuthorization)) and get back to step 3.
[...]

PR #1281, opened 2019-03-14
fix(otp-header): persist otp header on subsequent requests

ruxandrafed:
Related to #1279. Adds test to check for presence of otp header in requests made to list authorizations.


gr2m:
You have to add a `-test` suffix to the test in order for it to be run by the `npm test` script. Other than that the test looks 💯👍

ruxandrafed:
@gr2m Thanks for catching that! Fixed.

gr2m:
Authentication works using an internal plugin, it exists here:
https://github.com/octokit/rest.js/tree/master/plugins/authentication

plugins can hook into the request lifecycle using `octokit.hook.*`, in this case the authentication plugin runs code before a request is sent and in case of an error:

https://github.com/octokit/rest.js/blob/21e7a017366a3b029413d5ad51f34741bb0b59b8/plugins/authentication/index.js#L19-L20

In the error handler, we check if the error was due to a missing OTP, in which case we run your `auth.on2fa()` method and retry the request with the additional header

https://github.com/octokit/rest.js/blob/21e7a017366a3b029413d5ad51f34741bb0b59b8/plugins/authentication/request-error.js#L22-L31

Now what we want is not only to retry this request but also set the `x-github-otp` header for subsequent requests. 

So what I would probably try to do is use the `state` variable that is passed implicitly per pre-binding to both the before- and error-handler and store the otp on it. Then in the before handler check if it’s set and if it is, set the `x-github-otp` header.

That would be a good start I think :)

ruxandrafed:
Thanks @gr2m 😄 I actually just left a comment in the issue before I saw your reply. Pushed one more commit with the fix, it should pass the test, but I look forward to your review.

plugins/authentication/before-request.js
```diff
@@ -3,6 +3,10 @@ module.exports = authenticationBeforeRequest
 const btoa = require('btoa-lite')
 
 function authenticationBeforeRequest (state, options) {
+  if (state.otp) {
+    options.headers['x-github-otp'] = state.otp
+  }
+
```
gr2m: maybe check for `state.otp` below after we check for `state.auth.username`, as it’s only relevant in combination with Basic authentication?

plugins/authentication/request-error.js
```diff
@@ -27,6 +27,7 @@ function authenticationRequestError (state, error, options) {
       const newOptions = Object.assign(options, {
         headers: Object.assign({ 'x-github-otp': oneTimePassword }, options.headers)
       })
+      state.otp = oneTimePassword
```
gr2m, line 27: 👍

gr2m (commented):
Looks good, just one comment

plugins/authentication/before-request.js
ruxandrafed: Makes sense! Moved it below: https://github.com/octokit/rest.js/pull/1281/commits/856a7912786a75f7a9cf1b440cfd43ea5ac6b41e.

gr2m:
I wonder if we should update this condition here

https://github.com/octokit/rest.js/blob/21e7a017366a3b029413d5ad51f34741bb0b59b8/plugins/authentication/request-error.js#L14

It will fail if the set OTP becomes invalid. I think the better behaviour would be that it would ask for an OTP again as soon as it becomes invalid, but only if it was valid before. We could additionally check if `state.otp` is set? We would also need to unset it so we don’t up in an infinite loop the user keeps entering an invalid code.

Maybe something like this?

```js
  if (error.status === 401 && otpRequired && error.request && error.request.headers['x-github-otp']) {
    if (state.otp) {
      delete state.otp // no longer valid, request again
    } else {
      throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
    }
  }
```

gr2m:
^^ we can also do that in a follow up PR if you prefer, the current PR looks good as is

ruxandrafed:
@gr2m Sounds good, I'll do another PR! Thank you.

plugins/authentication/request-error.js
```diff
@@ -27,6 +27,7 @@ function authenticationRequestError (state, error, options) {
...
       return state.octokit.request(newOptions)
```
gr2m: I _think_ that with the addition to `before-request.js`, we no longer need to set the `x-github-otp` header here, we could remove lines 27-29 and just do `return state.octokit.request(options)`. I didn’t test it though.

I’ll be offline now and back tomorrow

ruxandrafed: You're right, it makes sense. Refactored in https://github.com/octokit/rest.js/pull/1281/commits/e635dc9d4b2df22ccd0e25f61a6eb9703a255854.

Will look at the other changes tomorrow. Have a good night!

gr2m (approved):
A+ pull request would merge again 👍 Thanks!

PR #1285, opened 2019-03-15
fix(otp-header): persist otp code only if valid

ruxandrafed:
Following up on https://github.com/octokit/rest.js/pull/1281#issuecomment-472710162, I followed your suggestions @gr2m but I must confess I'm not sure how to finish this. I started one way (https://github.com/ruxandrafed/rest.js/commit/5cda835952adcb1a45fe78c892db3f891de36495), but realized that's not exactly what you meant.

Now, I'm not sure how to fix this test and get back to full coverage 😬 Thanks for your patience!

### DIFF
--- a/plugins/authentication/request-error.js
+++ b/plugins/authentication/request-error.js
@@ -12,7 +12,11 @@ function authenticationRequestError (state, error, options) {
   }
 
   if (error.status === 401 && otpRequired && error.request && error.request.headers['x-github-otp']) {
-    throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
+    if (state.otp) {
+      delete state.otp // no longer valid, request again
+    } else {
+      throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
+    }
   }
 
   if (typeof state.auth.on2fa !== 'function') {
@@ -24,7 +28,14 @@ function authenticationRequestError (state, error, options) {
       return state.auth.on2fa()
     })
     .then((oneTimePassword) => {
-      state.otp = oneTimePassword
-      return state.octokit.request(options)
+      const newOptions = Object.assign(options, {
+        headers: Object.assign({ 'x-github-otp': oneTimePassword }, options.headers)
+      })
+      return state.octokit.request(newOptions)
+        .then(response => {
+          // If OTP still valid, then persist it for following requests
+          state.otp = oneTimePassword
+          return response
+        })
     })
 }
--- a/test/issues/1279-store-otp-send-header-all-requests-test.js
+++ b/test/issues/1279-store-otp-send-header-all-requests-test.js
@@ -46,4 +46,68 @@ describe('https://github.com/octokit/rest.js/issues/1279', () => {
         return octokit.authorization.listAuthorizations({ per_page: 100 })
       })
   })
+
+  it('prompts for OTP again once OTP code becomes invalid', () => {
+    nock('https://authentication-test-host.com', {
+      reqheaders: {
+        authorization: 'Basic dXNlcm5hbWU6cGFzc3dvcmQ='
+      }
+    })
+      .get('/')
+      .reply(401, {}, {
+        'x-github-otp': 'required; app'
+      })
+
+    nock('https://authentication-test-host.com', {
+      reqheaders: {
+        authorization: 'Basic dXNlcm5hbWU6cGFzc3dvcmQ=',
+        'x-github-otp': '123456'
+      }
+    })
+      .get('/')
+      .reply(200, {})
+
+    // OTP code now becomes invalid
+    nock('https://authentication-test-host.com', {
+      reqheaders: {
+        authorization: 'Basic dXNlcm5hbWU6cGFzc3dvcmQ=',
+        'x-github-otp': '123456'
+      }
+    })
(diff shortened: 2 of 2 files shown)


plugins/authentication/request-error.js
```diff
@@ -12,7 +12,11 @@ function authenticationRequestError (state, error, options) {
   }
 
   if (error.status === 401 && otpRequired && error.request && error.request.headers['x-github-otp']) {
-    throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
+    if (state.otp) {
+      delete state.otp // no longer valid, request again
+    } else {
+      throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
+    }
```
gr2m, line 19: ### 🤔 

This is more tricky than I thought. Right now we set `state.otp` right after `on2fa()` was resolved, we don’t know at this point if the OTP is valid or not. If it’s invalid, then the "no longer valid" is incorrect, it was never valid before. This is why the "basic with 2fa and invalid one-time-password" test is currently failing, a 3rd request is being sent because the code above no longer throws an error, instead it deletes `state.otp` and then sets it again from the `on2fa` method.

I _think_ what the correct thing to do would be to set `state.otp = oneTimePassword` after `state.octokit.request(options)` resolved

```diff
     .then((oneTimePassword) => {
-      state.otp = oneTimePassword
       return state.octokit.request(options)
+        .then(response => {
+          state.otp = oneTimePassword
+          return response
+        })
     })
```

Now that `state.otp` is set after `state.octokit.request(options)` we would need to set the OTP header for this one request manually. 

The benefit would be that the request would fail right away if the OTP is incorrect. If `state.otp` is set we can be sure that it was valid before and run the `auth.on2fa` function again.

Does that make sense?

gr2m:
@ruxandrafed do you plan to continue working on this? No rush, just checking in. And no worries if you can’t, just let me know :)

ruxandrafed:
@gr2m Yes, of course! Just fought with the flu over the last few days and caught up with work. Will get back to this ASAP 😄 Thanks!

ruxandrafed:
Okay, finally got back to this! Here's what I think makes sense, @gr2m .. lemme know if you agree, pls 😄 

- Once we get the OTP, we still need to pass it as a header, so I had to restore that part.
- If the request is successful, then we know the OTP is valid, hence we'll persist it in state
- For subsequent requests, if we found it persisted, it means it was valid last time we tried it, so we try using it again.

```diff
--- plugins/authentication/request-error.js
+++ plugins/authentication/request-error.js
@@ -12,11 +12,7 @@
   }
 
   if (error.status === 401 && otpRequired && error.request && error.request.headers['x-github-otp']) {
-    if (state.otp) {
-      delete state.otp // no longer valid, request again
-    } else {
-      throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
-    }
+    throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
   }
 
   if (typeof state.auth.on2fa !== 'function') {
@@ -28,7 +24,14 @@
       return state.auth.on2fa()
     })
     .then((oneTimePassword) => {
-      state.otp = oneTimePassword
-      return state.octokit.request(options)
+      const newOptions = Object.assign(options, {
+        headers: Object.assign({ 'x-github-otp': oneTimePassword }, options.headers)
+      })
+      return state.octokit.request(newOptions)
+        .then(response => {
+          // If OTP still valid, then persist it for following requests
+          state.otp = oneTimePassword
+          return response
+        })
     })
 }
```

I think this should be fine now, right? Thanks for your patience!

plugins/authentication/request-error.js
```diff
@@ -12,11 +12,7 @@ function authenticationRequestError (state, error, options) {
   }
 
   if (error.status === 401 && otpRequired && error.request && error.request.headers['x-github-otp']) {
-    if (state.otp) {
-      delete state.otp // no longer valid, request again
-    } else {
-      throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
-    }
+    throw new HttpError('Invalid one-time password for two-factor authentication', 401, error.headers, options)
```
gr2m: I _think_ we should leave the code above as it was.

When `state.otp` is set, it means that it was valid at least once. If we get a 401 now due to an invalid OTP then we should call `auth.on2fa` again, so we shouldn’t throw an error, only delete the now invalid `state.otp`.

gr2m (changes requested):
Can you add a test which sends a request successfully and then send another request which returns an error simulating that the OTP is no longer valid? In that case it should run the `on2fa()` method again

plugins/authentication/request-error.js
ruxandrafed: You're right, thanks!

ruxandrafed:
@gr2m Reviewed and added a test for this scenario. Does it look like a solid way to assert that we're calling `on2fa` again to you? Thanks!

gr2m (approved):
This is perfect, really well done, than you Ruxandra! 

I just took the liberty to squash the commits and push the test commit first to show that the tests where failing before the fix, but didn’t change anything else. I made sure to keep you as the author of the commits, too, of course.

ruxandrafed:
Thank you so much for your patience, @gr2m!

PR #1289, opened 2019-03-19
normalize pagination results beyond search

### LINKED ISSUE #1283: normalize pagination
I thought we did the normalization already via https://github.com/octokit/rest.js/commit/8373934ed21eb69c8dbacc3e06e1dc1566a6a7d8 but turns out that the name of the `key` which holds the search results is not always `items`, but depends on the endpoint. So while search results are normalized correctly, the list of check runs, check suits and installation repositories is not

- https://developer.github.com/v3/search/#example (key `items`)
- https://developer.github.com/v3/checks/runs/#response-3 (key: `check_runs`)
- https://developer.github.com/v3/checks/suites/#response-1 (key: `check_suites`)
- https://developer.github.com/v3/apps/installations/#list-repositories (key: `repositories`)

Although it’s a bug because it’s documented differently, the likely hood that fixing it might break things for people is high enough that we should publish this as breaking change out of caution.

---

Make sure to not accidentally include [Get the combined status for a specific ref](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref), it looks similar to the affected endpoints above, but does not actually paginate

gr2m:
closes #1283

- [x] normalize paginated results of all endpoints that use a namespace
- [x] deprecate `.total_count`, `.incomplete_results` and `.<namespace>`
- [x] make sure that other responses with `.total_count` in response do not paginate


tcbyrd:
Makes sense to me! Are you thinking nobody should need `total_count` here since you could just call `.length` on the results?

gr2m:
> Makes sense to me! Are you thinking nobody should need `total_count` here since you could just call `.length` on the results?

`total_count` and `data.length` can be different, if there are 100 results total but the current response only returns 30, then `data.length` would be 30, while `total_count` would say 100

Issue #1290, opened 2019-03-19
Document 2fa sms requirements 


karaggeorge:
More of a documentation request. It took me a while to figure out why I wasn't getting a 2fa code via sms when authenticating using the username/password method. After looking for a while I discovered this comment: https://github.com/sigmavirus24/github3.py/issues/387#issuecomment-110902328

The sms code is not sent until you actually try to create a token (or call `oauthAuthorizations.createAuthorization` of the API), even though the passed 2fa method gets called every time. After calling that API method, I finally get an sms, and resolving 2fa with that does the trick. 

I think it would be nice to note that somewhere in the docs

gr2m:
I didn’t know that, thanks for sharing! I’ll give this some thought. I wonder if we could do some kind of trick to make sure that an SMS is always sent if 2FA is enabled for basic authentication 🤔

gr2m:
I think one trick could be to actually create a token, but then delete it again right after (if 2FA is not enabled) before making the actual first request.

I also plan to add options to [`auth.*`](https://github.com/octokit/rest.js/tree/b273efe41745a8222d5093d58243298bc0b0b3e3#authentication) to let people create a token that will be used for further requests instead of using username & password, that should mitigate the problem.

karaggeorge:
I think the docs basically say the the username/password/2fa method of login is supposed to be very short-term, specifically used to create an actual token. That's why they only send the sms when you try to create a token. With that in mind, I think the second method sounds better. So, even if they do username/password/2fa, it could generate a token (maybe also request a `note` param for the name of the token) and then use that for the rest of the calls.

I'd love to open a PR for it if you direct me to where I should make the change. Looking at the code, it looks like most of the auth happens when the first request is made, rather than when the instance is created

EDIT: I actually couldn't find anything in their docs about it, other than the comment I linked in the description which seems to be coming from their support team describing the expected behavior. GitHub should probably include that in their API documentation as well.

gr2m:
> EDIT: I actually couldn't find anything in their docs about it, other than the comment I linked in the description which seems to be coming from their support team describing the expected behavior. GitHub should probably include that in their API documentation as well.

I’ve shared that with the docs team, they are already aware of lack of documentation around 2FA and SMS.

> I'd love to open a PR for it if you direct me to where I should make the change. Looking at the code, it looks like most of the auth happens when the first request is made, rather than when the instance is created

[...]

PR #1293, opened 2019-03-21
Update auth with 2fa

karaggeorge:
WIP for #1290 

-----
[View rendered README.md](https://github.com/karaggeorge/rest.js/blob/update-2fa-auth/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,16 +151,20 @@ The `auth` option can be
    - GitHub App bearer tokens
    - GitHub App installation tokens
 
-2. As object with the properties `username`, `password`, `on2fa`.
+2. As object with the properties `username`, `password`, `on2fa`, `note`, `scopes`.
 
    `on2fa` is an asynchronous function that must resolve with two-factor
    authentication code sent to the user.
+
+   `note` and `scopes` are used in the [creation of a token](https://octokit.github.io/rest.js/#api-OauthAuthorizations-createAuthorization).
 
    ```js
    new Octokit({
      auth: {
        username: 'octocat',
        password: 'secret',
+       note: 'my-token',
+       scopes: ['repo'],
        async on2fa () {
          // example: ask the user
          return prompt('Two-factor authentication Code:')


gr2m:
I talked with someone at GitHub and the road forward will be that we want to enforce the usage of tokens, so the options for the token note and required scopes will become mandatory in future and we will log a warning if it’s not set. That way we also get around the problem with SMS not being sent.

I think we should add to the documentation that when using basic auth, a token is created / reset using the given note and that that token is used for subsequent requests.

I’m not yet sure about the `note` option. It makes sense in the context of creating tokens, but in the context of basic authentication we might find a better name for that option that is more clear without the need of an explanation?

karaggeorge:
Well, I think that the user should know, like you said, that this auth method basically creates a token up front and then uses token auth for the rest of the calls. So, in that case, it does make sense to have `note`/`scopes` for that token. I still think the user needs some kind of way to receive the token for later use, maybe a `onTokenGenerated(token)` call you pass in the auth options that will get called with the token?

For the `note` and `scopes` options, I agree that a message if you don't pass them in would be nice, or default values, like the project's name from `package.json` for `note` and `[]` for scopes (since it is optional in creating tokens anyway).

gr2m:
> For the `note` and `scopes` options, I agree that a message if you don't pass them in would be nice, or default values, like the project's name from `package.json` for `note` and `[]` for scopes (since it is optional in creating tokens anyway).

I wouldn’t set default values for `note` and `scope`, at least not without releasing a new breaking version, as it might break people’s code. Basic authentication allows for everything, `[]` allows for nearly nothing. If anything we could default `scopes` to all the scopes there are, this would be a non-breaking change, but I don’t think that it would be  a good idea 🤔 

I think a warning is a better idea for now, then with the next breaking version, I would throw an error with `note` is not set. We cannot access `package.json` because Octokit is meant to be run browsers, too.

README.md
```diff
@@ -151,16 +151,20 @@ The `auth` option can be
    - GitHub App bearer tokens
    - GitHub App installation tokens
 
-2. As object with the properties `username`, `password`, `on2fa`.
+2. As object with the properties `username`, `password`, `on2fa`, `note`, `scopes`.
```
gr2m, line 154: I think instead of adding `note` and `scopes`, we should add a `token` option which has the same properties as the [Create a new Authorization](https://developer.github.com/v3/oauth_authorizations/#parameters), but in camelCase:

```js
new Octokit({
  auth: {
    username: 'octocat',
    password: 'secret',
    token: {
      note: 'my-token',
      scopes: ['repo'],
      // defaults to documentation URL explaining why tokens are created with Basic auth in Octokit,
      noteUrl: '',
      // Optional: Allow to create a token for an OAuth app
      clientId: 123,
      clientSecret: 'secret',
      fingerprint: 'fingerprint'
    },
    async on2fa () {
      // example: ask the user
      return prompt('Two-factor authentication Code:')
    }
  }
})
```

If `token` is not set, we will start logging a warning with a URL to documentation which explains why, same as the `token.noteUrl` default

README.md
```diff
@@ -151,16 +151,20 @@ The `auth` option can be
...
 
    `on2fa` is an asynchronous function that must resolve with two-factor
    authentication code sent to the user.
 
+   `note` and `scopes` are used in the [creation of a token](https://octokit.github.io/rest.js/#api-OauthAuthorizations-createAuthorization).
```
gr2m, line 159: I think I would remove this paragraph and instead put a paragraph on top of the section explaining that Basic Authentication is used to create a token, which in turn is used for subsequent requests.

gr2m:
Here is what I plan to implement. 

API:

```js
new Octokit({
  auth: {
    username: 'octocat',
    password: 'secret',
    token: {
      // defaults to "octokit <timestamp> <fingerprint>", where timestamp is YYYY-MM-DD and fingerprint is a random string
      // This will prevent conflicts when multiple tokens are created
      note: 'octokit 2019-04-03 abc4567',
      // defaults to empty array
      scopes: [],
      // defaults to documentation URL explaining why tokens are created with Basic auth in Octokit,
      noteUrl: 'https://github.com/octokit/rest.js#authentication',
      // defaults to random string
      fingerprint: 'abc4567'
      // Optional: Allow to create a token for an OAuth app
      clientId: 123,
      clientSecret: 'secret',
    },
    async on2fa () {
      // example: ask the user
      return prompt('Two-factor authentication Code:')
    }
  }
})
```

The `auth.token` option is optional in order to not introduce a breaking change. A warning is logged when `auth.token` is not set, explaining that in case the user has 2FA + SMS enabled, they will not receive an SMS until a request to create a token is sent.

With the next breaking version, the `auth.token.*` options will be set implicitly to the documented defaults.

This will be accompanied by a new API to retrieve the current authentication, see https://github.com/octokit/rest.js/issues/1298#issuecomment-479614921. The retrieved authentication will include the token id which can be used to delete an authorization

karaggeorge:
I think this sounds like a solid plan. Do you want me to start implementing any of the above?

gr2m:
Hold on, I’m working on this right now.

gr2m:
I’ll finish this up once soon

gr2m:
@karaggeorge I’m sorry it took sooo long, but I think the authentication bit is done now. It took so long because I ended up moving all the authentication logic into its own packages.

The way it will work is this:

```js
const { createBasicAuth } = require("@octokit/auth");
const Octokit = require("@octokit/auth");

const auth = createBasicAuth({
  username: 'octocat',
  password: 'secret',
  async on2fa () {
    // example: ask the user
    return prompt('Two-factor authentication Code:')
  },
  // "token" is optional
  token: {
    // defaults to "octokit <timestamp> <fingerprint>", where timestamp is YYYY-MM-DD and fingerprint is a random string
    // This will prevent conflicts when multiple tokens are created
    note: 'octokit 2019-04-03 abc4567',
    // defaults to empty array
    scopes: [],
    // defaults to documentation URL explaining why tokens are created with Basic auth in Octokit,
    noteUrl: 'https://github.com/octokit/rest.js#authentication',
    // defaults to random string unless token.note is set
    fingerprint: 'abc4567'
    // create a token for an OAuth app
    clientId: 123,
    clientSecret: 'secret',
  }
})

const octokit = new Octokit({ auth })
```

This will hook into the request lifecycle using `octokit.hook` and set the required authentication headers based on the requested URL: basic auth for URLs matching

- `/authorizations`
- `/authorizations/clients/:client_id`
- `/authorizations/clients/:client_id/:fingerprint`
- `/authorizations/:authorization_id`

otherwise a token is created and used.

It will handle 2Fa authentication, including making sure that the TOTP is send if SMS is configured for delivery. 

If you want to get the token or the TOTP you an use

```js
const { token } = await auth({ type: 'token' })
const { totp } = await auth({ type: 'basic' })
```

I would super appreciate if you could look through the docs of the upcoming package: https://github.com/octokit/auth-basic.js/tree/1/initial-version

and if you could give it a try, that would be fantastic. You’ll need to run `npm run build` and then `require('/path/to/octokit/auth-basic.js/pkg')`

You can directly comment on the pull request for the initial version: https://github.com/octokit/auth-basic.js/pull/1

README.md
```diff
@@ -151,16 +151,20 @@ The `auth` option can be
    - GitHub App bearer tokens
    - GitHub App installation tokens
```
ruzyysmartt:
```suggestion
   - GitHub App installation tokens
```

PR #1304, opened 2019-04-04
.paginate() with results namespace (GET /installation/repositories, single page response)

gr2m:


gr2m:
@tcbyrd this should fix the pagination problem you run into with https://github.com/probot/scheduler, could you check?

PR #1310, opened 2019-04-06
fix(2fa): avoid caching bad 2fa token

### LINKED ISSUE #1309: Expired OTP is inadvertently cached
When an OTP expires, octokit generates a new one via `on2fa` but doesn't use it.

The ordering of arguments passed to `Object.assign` at https://github.com/octokit/rest.js/blob/eb1ebab975cb3a5877e7df25acc500b3e048b49e/plugins/authentication/request-error.js#L32 means that even though it calls the `on2fa()` function and generates a new OTP, the otp already in `options.headers` is the one that ultimately gets used.

Reversing the order of arguments fixes the problem for me. I'm happy to open a PR if that's the preferred fix, but I'm not very familiar with the codebase and I wonder if it's more appropriate to either invalidate something in `state` or delete the header.

ianwremmel:
closes #1309

### DIFF
--- a/test/integration/authentication-test.js
+++ b/test/integration/authentication-test.js
@@ -202,8 +202,7 @@ describe('authentication', () => {
         username: 'username',
         password: 'password',
         on2fa () {
-          callCount++
-          return callCount
+          return ++callCount
         }
       }
     })
--- a/test/integration/authentication-test.js
+++ b/test/integration/authentication-test.js
@@ -155,6 +155,66 @@ describe('authentication', () => {
       })
   })
 
+  it('basic with expiring 2fa', () => {
+    nock('https://authentication-test-host.com', {
+      reqheaders: {
+        authorization: 'Basic dXNlcm5hbWU6cGFzc3dvcmQ='
+      }
+    })
+      .get('/')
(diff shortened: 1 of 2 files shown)


test/integration/authentication-test.js
```diff
@@ -155,6 +155,66 @@ describe('authentication', () => {
...
+
+    let callCount = 0
+    const octokit = new Octokit({
+      baseUrl: 'https://authentication-test-host.com',
+      auth: {
+        username: 'username',
+        password: 'password',
+        on2fa () {
+          callCount++
+          return callCount
```
gr2m: you can replace the two lines above with

```js
return ++callCount
```

gr2m (changes requested):
This looks great, just a tiny code simplification. I can fix that up, too, just let me know :)

test/integration/authentication-test.js
ianwremmel: changed

gr2m (approved):
Thanks a lot Ian 💝

gr2m:
Also thanks @ruxandrafed for your review 👍

PR #1320, opened 2019-04-14
add type for basic auth

protoEvangelion:
Fixes error: 

```
node_modules/@octokit/rest/index.d.ts:47:5
    47     auth?:
           ~~~~
    The expected type comes from property 'auth' which is declared here on type 'Options'
```

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -45,7 +45,7 @@ declare namespace Octokit {
   }
 
   export interface Options {
-    auth?: string | { username: string; password: string; on2fa: () => Promise<string> } | { clientId: string; clientSecret: string; } | { (): (string | Promise<string>) };
+    auth?: string | { username: string; password: string; } | { username: string; password: string; on2fa: () => Promise<string> } | { clientId: string; clientSecret: string; } | { (): (string | Promise<string>) };
     userAgent?: string;
     previews?: string[];
     baseUrl?: string;


scripts/templates/index.d.ts.tpl
```diff
@@ -45,7 +45,7 @@ declare namespace Octokit {
   }
 
   export interface Options {
-    auth?: string | { username: string; password: string; on2fa: () => Promise<string> } | { clientId: string; clientSecret: string; } | { (): (string | Promise<string>) };
+    auth?: string | { username: string; password: string; } | { username: string; password: string; on2fa: () => Promise<string> } | { clientId: string; clientSecret: string; } | { (): (string | Promise<string>) };
```
gr2m, line 48: I’m not sure ... I think you should always set the on2fa method, we should probably throw an error if it’s not set. You can just make it a no-op if you know what you are doing. But allowing people to just set username/password potentially without understanding the implications might be a risk for them to shoot in their foot?

protoEvangelion: OK so I have two cases that I need to handle:
1. user has 2fa enabled
2. user has 2fa disabled

Are you saying I can pass the `on2fa` function in either case and it will handle it for me?

gr2m: Yes, it will only be used if the user has 2fa enabled

protoEvangelion: Even better! Thanks for clarifying. No need for this then.

Issue #1329, opened 2019-04-15
new website: follow up issues


gr2m:
https://octokit.github.io/rest.js/

- [ ] whitespace on right of header bar for screens < 360px  (https://github.com/gr2m/octokit-rest-documentation/issues/42)
- [x] After click on menu, target heading is hidden behind header bar (https://github.com/gr2m/octokit-rest-documentation/issues/40)
- [ ] Update URL when scrolling between sections (https://github.com/gr2m/octokit-rest-documentation/issues/5)
- [ ] Make 3 column design (https://github.com/gr2m/octokit-rest-documentation/issues/20)

janl:
doesn’t work in Safari/Mac:

```
[Error] ReferenceError: Can't find variable: IntersectionObserver
	da (app-84d26d92f63eeec6c7e1.js:1:134948)
	callback (app-84d26d92f63eeec6c7e1.js:1:140341)
	oa (app-84d26d92f63eeec6c7e1.js:1:132784)
	ra (app-84d26d92f63eeec6c7e1.js:1:132532)
	Ba (app-84d26d92f63eeec6c7e1.js:1:142546)
	Va (app-84d26d92f63eeec6c7e1.js:1:145020)
	(anonymous function) (app-84d26d92f63eeec6c7e1.js:1:158227)
	unstable_runWithPriority (app-84d26d92f63eeec6c7e1.js:1:168940)
	jl (app-84d26d92f63eeec6c7e1.js:1:158184)
	Ol (app-84d26d92f63eeec6c7e1.js:1:157938)
	Cl (app-84d26d92f63eeec6c7e1.js:1:157283)
	kl (app-84d26d92f63eeec6c7e1.js:1:156304)
	Ja (app-84d26d92f63eeec6c7e1.js:1:155178)
	Il (app-84d26d92f63eeec6c7e1.js:1:159193)
	render (app-84d26d92f63eeec6c7e1.js:1:163038)
	(anonymous function) (app-84d26d92f63eeec6c7e1.js:1:160897)
	Hl (app-84d26d92f63eeec6c7e1.js:1:160826)
	(anonymous function) (app-84d26d92f63eeec6c7e1.js:1:190760)
[Error] ReferenceError: Can't find variable: IntersectionObserver
	da (app-84d26d92f63eeec6c7e1.js:1:134948)
	callback (app-84d26d92f63eeec6c7e1.js:1:139995)
	oa (app-84d26d92f63eeec6c7e1.js:1:132784)
[...]

PR #1349, opened 2019-04-30
make use of IntersectionObserver optional

gr2m:
addresses https://github.com/octokit/rest.js/issues/1329#issuecomment-487873471


IxxxUMe:
https://github.com/octokit/rest.js/pull/1349#issuecomment-488132480

PR #1352, opened 2019-05-01
documentation: usage example updates

gr2m:
This will look much better once the code is on the side

-----
[View rendered docs/src/pages/api/00_usage.md](https://github.com/octokit/rest.js/blob/docs/usage-example/docs/src/pages/api/00_usage.md)
[View rendered docs/src/pages/api/03_request_formats.md](https://github.com/octokit/rest.js/blob/docs/usage-example/docs/src/pages/api/03_request_formats.md)

### DIFF
--- a/docs/src/pages/api/00_usage.md
+++ b/docs/src/pages/api/00_usage.md
@@ -2,12 +2,191 @@
 title: "Usage"
 ---
 
-### Node
+Import the Octokit constructor.
 
-Install with `npm install @octokit/rest`.
+For direct usage in browsers, download `octokit-rest.min.js` from the [latest release](https://github.com/octokit/rest.js/releases/latest) and import it using a `<script src="octokit-rest.min.js"></script>` tag.
 
 ```js
-const octokit = require('@octokit/rest')()
+const Octokit = require('@octokit/rest')
+```
+
+Now instantiate your octokit API. All options are optional, but authentication is strongly encouraged.
+
+```js
+const octokit = Octokit({
+```
+
+You can set `auth` to a personal access token string.
+
+Learn more about [authentication](#authentication).
+
+```js
+  auth: 'secret123'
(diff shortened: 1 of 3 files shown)


docs/src/pages/api/00_usage.md
```diff
@@ -2,86 +2,197 @@
...
 
-### Node
+Import the Octokit constructor.
 
-Install with `npm install @octokit/rest`.
+For direct usage in browsers, download `octokit-rest.min.js` from the [latest release](https://github.com/octokit/rest.js/releases/latest) and import it using a `<script src="octokit-rest.min.js"></script>` tag.
 
 ```js
-const octokit = require('@octokit/rest')()
+const Octokit = require('@octokit/rest')
```
BarbourSmith, line 10: Is it also possible to use the `import Octokit from 'octokit/rest'`? style of import?

docs/src/pages/api/00_usage.md
```diff
@@ -2,86 +2,197 @@
 title: "Usage"
 ---
 
-### Node
+Import the Octokit constructor.
 
-Install with `npm install @octokit/rest`.
+For direct usage in browsers, download `octokit-rest.min.js` from the [latest release](https://github.com/octokit/rest.js/releases/latest) and import it using a `<script src="octokit-rest.min.js"></script>` tag.
```
BarbourSmith, line 7: Why the change in focus from using npm to using the download in browsers?

gr2m: there is no change of focus, it’s a universal library that should run in all environments. I just updated the documentation with the content from the README.md

docs/src/pages/api/00_usage.md
eine: I'm also interested on this.

PR #1365, opened 2019-05-15
Update README.md

bradparks:
This simply adds how to install it to the readme, which is not necessarily obvious to people that don't use node a ton!

-----
[View rendered README.md](https://github.com/bradparks/rest.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@
 [![Greenkeeper](https://badges.greenkeeper.io/octokit/rest.js.svg)](https://greenkeeper.io/)
 
 ## Installation
-```$ npm install --save @octokit/rest```
+```shell
+npm install @octokit/rest
+```
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,9 @@
 [![Coverage Status](https://coveralls.io/repos/github/octokit/rest.js/badge.svg)](https://coveralls.io/github/octokit/rest.js)
 [![Greenkeeper](https://badges.greenkeeper.io/octokit/rest.js.svg)](https://greenkeeper.io/)
 
+## Installation
+```$ npm install --save @octokit/rest```
```
gr2m: Could you please make it a code block? You’ll have to put the \`\`\` into their own lines.

Remove the `$`, it makes it easier to copy without it. Remove the `--save`, it’s no longer needed.

gr2m (changes requested):
Thanks for the pull request! I removed installation information recently when I updated the website: https://octokit.github.io/rest.js/. But we can leave it in the README so people can see it on first glance.

Just have a few change requests. Thanks!

README.md
wolfy1339: It's better to be explicit here, as some people disable that functionality. Plus it's needed for older versions of npm

gr2m: It’s default behaviour for all supported node/npm versions. People who disable that functionality know what they are doing. Also `--save` is now `--save-prod` as far as I know. I would leave it out.

bradparks: I think I made the changes....

gr2m (approved):
Thanks Brad!

PR #1367, opened 2019-05-16
Greenkeeper/@octokit/request 4.0.1

gr2m:
closes #1366


gr2m:
Here is the standalone module that replaces `@octokit/request/lib/http-error`: https://github.com/octokit/request-error.js

Issue #1377, opened 2019-05-25
Add headers to the type definition for requests 


billytrend:
The headers field here works but is not included in the type definitions. I have a few use cases for it, eg for request correlation

```
await github.repos.get({
  owner: "octocat",
  repo: "Hello-World",
  headers: { "x-correlation-id": Math.random() },
})
```

It would also be nice to have plugin support for requests like this so that we can say pass an options object to a request and handle it within a plugin.

```
await github.repos.get(
  {
    owner: "octocat",
    repo: "Hello-World",
  },
  { correlationId: Math.random() },
);
```

Let me know if this sounds good or if there is already a good way to do these things and I can look at contributing the changes required.

gr2m:
What's the `x-correlation-id` header? I don’t think it’s mention anywhere in the docs for GitHub’s API, so I’m not sure why type definitions for it would make sense?

> It would also be nice to have plugin support for requests like this so that we can say pass an options object to a request and handle it within a plugin.

You can set custom meta options on the request option, e.g.

```js
await github.repos.get(
  {
    owner: "octocat",
    repo: "Hello-World",
    request: {
      correlationId: Math.random()
    }
  }
);
```

Then in your plugin you can define a request hook that does something with it.

See

- https://octokit.github.io/rest.js/#plugins
- https://octokit.github.io/rest.js/#hooks

Hope that helps

billytrend:
Ah the header is just something that gets automatically picked up by our telemetry libraries. I also had a use case for sending custom cache control headers at one point. 

I think the request option is exactly what I'm looking for, thanks!! Not sure if it's in type definitions though but I'll check later.

gr2m:
I’ll close the issue for the time being as the original question is resolved I think

PR #1378, opened 2019-05-25
Add request options to type definitions

billytrend:
I need to fully check I picked the right type here but it looks like it would address #1377

### DIFF
--- a/scripts/templates/index.d.ts.tpl
+++ b/scripts/templates/index.d.ts.tpl
@@ -5,6 +5,7 @@
 /// <reference lib="esnext.asynciterable" />
 
 import * as http from 'http'
+import { OctokitRequestOptions } from "@octokit/request/dist-types/types";
 
 declare namespace Octokit {
   type json = any
@@ -37,6 +38,10 @@ declare namespace Octokit {
     }
 
     [Symbol.iterator](): Iterator<any>
+  }
+
+  export type Params<T> = T & {
+    request?: OctokitRequestOptions;
   }
 
   export type AnyResponse = Response<any>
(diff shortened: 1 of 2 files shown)


wolfy1339:
I would have personally tested this myself, but since `node-gyp` doesn't support VS 2019 and npm doesn't support NodeJS 12 yet, I cannot.
I'm going to leave this for other contributors

scripts/templates/index.d.ts.tpl
```diff
@@ -5,6 +5,7 @@
 /// <reference lib="esnext.asynciterable" />
 
 import * as http from 'http'
+import { OctokitRequestOptions } from "@octokit/request/dist-types/types";
```
gr2m, line 8: If we depend on internal APIs we have to pin the `@octokit/request` version in package.json

gr2m:
@billytrend can you share a code example that currently does not work with `@octokit/rest` but would work once your PR is merged?

wolfy1339:
@billytrend Are you still interested in taking up this PR?

wolfy1339:
@billytrend ping

gr2m:
I would suggest to just copy & paste the definitions for now, I did that in a few other libraries, too.

The proper fix would be to have one repository with all the type definitions, e.g. `octokit/common-types.ts`, at least for the type definitions that are shared across multiple libraries

ArtekArtek (approved):

gr2m:
https://github.com/octokit/types.ts/ is a thing now

gr2m:
I think this is obsolete now?

PR #1380, opened 2019-05-29
fix: make search results in gh-pages scrollable

soonoo:
Added some styles to make search results in [octokit.github.io/rest.js](octokit.github.io/rest.js) scrollable.  
I'm not familiar with grid layout so please let me know if there is a problem.


gr2m (approved):
I tested it locally and it works great, thank you so much!

Issue #1383, opened 2019-05-30
Debouncing search in documentation


soonoo:
It seems searching in gh-pages is little bit laggy.
How about debouncing search function or processing it in web worker?
(Will it be an overkill?)

I want to work on it if possible.

By the way, thank you for the nice software😺

gr2m:
> How about debouncing search function or processing it in web worker?

I would _love_ if the search would run in a webworker. But I don’t know how to make it work. We are using this Gatsby Plugin: https://github.com/gatsby-contrib/gatsby-plugin-elasticlunr-search

And here is the custom JavaScript code that does the UI:
https://github.com/octokit/rest.js/blob/master/docs/src/components/search.js

The heavy load is the initial creation of the search index:
https://github.com/octokit/rest.js/blob/ade2d00dd0642a4205ab3a7714106003db820f68/docs/src/components/search.js#L64

And the actual search:
https://github.com/octokit/rest.js/blob/ade2d00dd0642a4205ab3a7714106003db820f68/docs/src/components/search.js#L94-L96

SGrondin:
I'd start with debouncing. I suggest doing it manually, but if you think it's worth using a library, feel free to reach out to me on the [Probot Slack](https://probot-slackin.herokuapp.com/) so it's done using the same library that the [throttling plugin](https://github.com/octokit/plugin-throttling.js) uses because it's extremely important to keep the package size small.

Debouncing is likely to be enough, but it will depend on the CPU intensiveness of the full text search. Most of the time Webworkers are overkill because it takes more work just to copy the data in and out of the worker than to do the work. However you might be able to use [ownership transfer](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Passing_data_by_transferring_ownership_(transferable_objects)) to avoid this issue, and full text search is often fairly CPU intensive.

It's possible that only the initial [indexing](https://github.com/octokit/rest.js/blob/ade2d00dd0642a4205ab3a7714106003db820f68/docs/src/components/search.js#L64) will benefit from Web workers. Make sure to benchmark the gains for indexing and for searching separately.

But please start with debouncing because it might be enough. Good luck! Feel free to reach out to me if you have questions.

soonoo:
Actually I tried debounced version before opening this issue, but there was no such a big effect. 

After that I've found below `setState` causes lag:
[...]

PR #1387, opened 2019-06-01
docs: improve search performance

soonoo:
Related issue: #1383 

What I did:
- renders only part of the search results and render the rest when scroll reaches to the end.
- debounces rendering of search results.

elasticlunr's search function seems fast enough and what slows website was search component's `render` function. (About 50 ~ 300 items are rendered whenever user types)


docs/package.json
```diff
@@ -17,9 +17,11 @@
     "gatsby-source-filesystem": "^2.0.28",
     "gatsby-transformer-remark": "^2.3.8",
     "lodash": "^4.17.11",
+    "lodash.debounce": "^4.0.8",
```
zeke: looks like the full `lodash` library is already in the mix, so you probably don't need `lodash.debounce`

docs/src/components/search.js
```diff
@@ -85,15 +97,32 @@ export default class Search extends Component {
...
+      // Map over each ID and return the full document
+      .map(({ ref }) => this.index.documentStore.getDoc(ref))
+
+    let height = 0
+    let visibleResultsCount
+    // Calculate maximum count of results that fits in current viewport height
+    for(const [index, doc] of results.entries()) {
+      // 54, 26 are the height of search results item.
+      // They are unknonw until rendered. Is there a better way to set these values?
+      if(doc.type === 'API method') height += 54
```
zeke: There probably is a more dynamic way to do this, but if you end up keeping this implementation I would suggest changing these [magic numbers](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) into `const` declarations.

docs/package.json
soonoo: I didn't konw that, uninstalled lodash.debounce!

docs/src/components/search.js
soonoo: I think it gets a bit complicated with dynamic way so I just end up keeping numeric values. Please check commits below.

docs/package.json
gr2m: Oops sorry I missed that bit, I think we need the `lodash.debounce` separately. `lodash` is used only in built scripts, not in the frontend of the documentation. The `lodash.debounce` method is currently imported in `docs/src/components/search.js` so we have to keep it, even though it looks weird in package.json

soonoo: https://github.com/octokit/rest.js/blob/b8f2097ef7bf03ffde602a6ef866898ee025dbf3/docs/src/components/search.js#L2

I forgot to change import statement and pushed the changed code. Is this okay with just `lodash` dependency?

gr2m: I’m not sure if gatsby does proper tree shaking to remove unused code, but it should be okay. Maybe compare the download size of JS assets locally to what we currently have in production? I thenk you have to run `gatsby build` and then `gatsby serve`. I don’t think `gatsby develop` will do all the bundling and minification

soonoo: Above import statement imports whole `lodash` library so changed it to import only a debounce function.
https://github.com/octokit/rest.js/blob/2bb9a17bd380bf0eaf9f1bf9baeb7d40d9bf357d/docs/src/components/search.js#L2

without debounce (24.21KB transferred and file size is 74.02KB):
2019-06-04T15:57:49

with debouce:
2019-06-04T15:57:40

Is this affordable?

gr2m: There is probably use for optimization but that can be addressed in a future PR :)

gr2m (approved):
This is great, thanks a lot Soonwoo! The search feels _much_ better now!

PR #1393, opened 2019-06-09
Add newest reactions support

skjnldsv:
See https://developer.github.com/v3/reactions/#reaction-types

Eyes and rocket are now supported into the api :)

### DIFF
--- a/plugins/rest-api-endpoints/routes.json
+++ b/plugins/rest-api-endpoints/routes.json
@@ -10004,6 +10004,8 @@
             "reactions-thinking_face",
             "reactions-heart",
             "reactions-tada",
+            "reactions-rocket",
+            "reactions-eyes",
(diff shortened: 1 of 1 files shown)


gr2m:
Thanks for the pull request! Unfortunately I cannot merge it, because the routes.json file is generated and your changes would be overwritten the next time [@octokit/routes](https://github.com/octokit/routes) are updated.

This would need to be solved in https://github.com/octokit/routes, to be precise in https://github.com/octokit/routes/blob/61e5c052356a0fb74dab9b65fa928877e5e5e14e/lib/endpoint/tags/table.js#L195-L205

Could you send a pull request updating that file, please?

skjnldsv:
Followup https://github.com/octokit/routes/pull/433 :)

PR #1402, opened 2019-06-16
Deprecation warning text edits for request options

bkoski:
This fixes a small typo in the deprecation warnings and fixes the link in these warnings to the request.js docs.

Bumped into this while doing a rest.js upgrade - the error was a bit puzzling so thought I would fix for others.


gr2m (approved):
This is great, thanks so much Ben1

PR #1404, opened 2019-06-17
Milestone Update: Allow due date to removed from a milestone

nwmac:
Currently there is no way with updateMilestone to remove the due date. This change allows passing null as the due_on parameter, which will remove the exsiting due date from the milestone.


gr2m:
Thanks for opening up the pull request. The `routes.json` file is generated, we cannot edit it directly, the changes will get overwritten the next time there is an update to https://github.com/octokit/routes.

The problem has to be fixed at https://github.com/octokit/routes. If you can, could you create an issue there?

PR #1407, opened 2019-06-18
Add a header for listing and replacing topics on a repository

jenniferemshepherd:
When making a request to list/replace topics, an error is shown:

> To view the topics property in calls that return repository results, you must provide a custom media type in the Accept header: 
> `application/vnd.github.mercy-preview+json`

as per the [docs](https://developer.github.com/v3/repos/#list-all-topics-for-a-repository).
This commit adds the header that is required for both methods.

### DIFF
--- a/plugins/rest-api-endpoints/routes.json
+++ b/plugins/rest-api-endpoints/routes.json
@@ -9029,6 +9029,9 @@
       "url": "/repos/:owner/:repo/teams"
     },
     "listTopics": {
+      "headers": {
+        "accept": "application/vnd.github.mercy-preview+json"
+      },
       "method": "GET",
       "params": {
         "owner": {
@@ -9371,6 +9374,9 @@
       "url": "/repos/:owner/:repo/branches/:branch/protection/restrictions/users"
     },
     "replaceTopics": {
+      "headers": {
+        "accept": "application/vnd.github.mercy-preview+json"
+      },
       "method": "PUT",
       "params": {
         "names": {


wolfy1339:
I believe this file is auto-generated, and is created by the data in the `octokit/routes` repository.

@gr2m will be able to confirm for surr

jenniferemshepherd:
Thanks - I can see this has been raised/referenced [here](https://github.com/octokit/routes/issues/356) and I've confirmed that this file *will* get overwritten by anything in `octokit/routes`.

Looks like these are the places it is needed: 

- https://github.com/octokit/routes/blob/d720612ace3457aa55f43affd1a4c858717fdc44/routes/api.github.com/repos/replace-topics.json
-https://github.com/octokit/routes/blob/d720612ace3457aa55f43affd1a4c858717fdc44/routes/api.github.com/repos/list-topics.json

Very happy to suggest changes over on that repo if those will then get reflected here!

wolfy1339:
Need to confirm if this was fixed in `@octokit/routes`

gr2m:
This is not yet fixed in `octokit/routes`:
https://github.com/octokit/routes/blob/261c763e51ea4d327fedc51cde182d2ffec65708/openapi/api.github.com/operations/repos/list-topics.json#L13-L21

Note that we migrated octokit/routes form our made-up JSON format to OpenAPI and we'll need a moment to upgrade our scripts to work with the new format.

For the time being, you can set custom preview headers like this:

```js
octokit.repos.listTopics({
  owner,
  repo,
  previews: ['mercy']
})
```

PR #1410, opened 2019-06-21
Add  checkVulnerabilityAlerts enpoint

### LINKED ISSUE #1409: Add check vulnerability alert enpoint
Currently, the endpoint for checking if vulnerability alerts are enabled on a repository, is not included in this library. 

I think it would be useful to add it, because the following endpoint `enableVulnerabilityAlerts` returns a `204` whether the vulnerability alerts are already enabled or disabled for a given repo.

[https://developer.github.com/v3/repos/#check-if-vulnerability-alerts-are-enabled-for-a-repository](https://developer.github.com/v3/repos/#check-if-vulnerability-alerts-are-enabled-for-a-repository)

manuelpizani:
Closes #1409

### DIFF
--- a/plugins/rest-api-endpoints/routes.json
+++ b/plugins/rest-api-endpoints/routes.json
@@ -7572,6 +7572,23 @@
       },
       "url": "/repos/:owner/:repo/pages"
     },
+    "checkVulnerabilityAlerts": {
+      "headers": {
+        "accept": "application/vnd.github.dorian-preview+json"
+      },
+      "method": "GET",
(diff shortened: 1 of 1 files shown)


gr2m:
There is a problem in https://github.com/octokit/routes that I hope I’ll be able to resolve today. Once fixed, the new endpoint will be added, along with some other updates. The `routes.json` file is generated each time `@octokit/routes` has an update, we do not edit it manually. We can leave #1409 open for tracking

PR #1414, opened 2019-06-26
Update App example in Authentication docs

nason:
I ran into a module import error when I followed these instructions. `App` should be a named export here:) https://github.com/octokit/app.js

-----
[View rendered docs/src/pages/api/01_authentication.md](https://github.com/nason/rest.js/blob/patch-1/docs/src/pages/api/01_authentication.md)


gr2m (approved):
It should, thanks so much!

PR #1427, opened 2019-07-17
Added docs for advanced requests using AbortController

### LINKED ISSUE #1426: Abortable Request Support
Hi,
I am trying to add a `signal` attribute as we can do with `Abortable Fetch` API.

I checked the docs for rest.js but could not find anything that I can use.
Any help would be appreciated.

inaseem:
**What's new?**
- fixes #1426 
- Added documentation for `advanced request formats` that can be aborted by a user when desired.
- Changes to [Request formats](https://github.com/octokit/rest.js/blob/master/docs/src/pages/api/03_request_formats.md) are shown below.

### Advanced Requests
The [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) interface can be used to abort one or more requests as and when desired.
When the request is initiated, we can pass in the [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) as an option inside the request's options object.

```js
const controller = new AbortController()
const signal = controller.signal
const { data: prDiff } = await octokit.pulls.get({
  owner: 'octokit',
  repo: 'rest.js',
  pull_number: 1278,
  mediaType: {
    format: 'diff'
  },
  request: {
    signal
  }
})
```

Use `controller.abort()` to abort the request when desired.

-----
[View rendered docs/src/pages/api/03_request_formats.md](https://github.com/naseemali925/rest.js/blob/docs/abortable-request/docs/src/pages/api/03_request_formats.md)

### DIFF
--- a/docs/src/pages/api/03_request_formats.md
+++ b/docs/src/pages/api/03_request_formats.md
@@ -1,5 +1,5 @@
 ---
-title: "Request formats"
+title: "Request formats & aborts"
 ---
 
 Some API endpoints support alternative response formats, see [Media types](https://developer.github.com/v3/media/).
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thanks a bunch 💐 I made a few minor changes in e745f29564ea2f1e614ba521091c66ab9192dfce but otherwise this is great

PR #1457, opened 2019-08-29
fix: remove environment enum

ghost:
This can actually be any valid string, such as "development". The enum was taken from the deployments documentation, which lists `production`, `staging` and `qa` as examples, but does not state they are the only values allowed.

### DIFF
--- a/plugins/rest-api-endpoints/routes.json
+++ b/plugins/rest-api-endpoints/routes.json
@@ -7004,11 +7004,6 @@
           "type": "string"
         },
         "environment": {
-          "enum": [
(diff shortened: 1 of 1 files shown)


gr2m:
the `routes.json` file is generated with each update of https://github.com/octokit/routes/. We could look for the `e.g.` before the list of values and in that case do not turn the value into an enum

https://developer.github.com/v3/repos/deployments/#parameters-1

Can you create an issue in the `octokit/routes` repository and we take from there?

gr2m:
this has been resolved

PR #1481, opened 2019-09-23
fix(elasticlunr-plugin): use url field instead of url

### LINKED ISSUE #1475: docs: search typeahead shows "undefined" for endpoint URL and does not show the method name
The first item should show `GET /repos/:owner/:repo/git/refs/:ref` and `octokit.git.getRef()`

image

This must have been caused by #1466

copperwall:
This adds a fix in the elasticlunr search plugin resolvers config to use
the url field instead of path when building an index of search results.
The url after the method in search results was undefined, now it's the
correct API URL for that resource.

This also updates the method index field to use OctokitApiMethod.example
to show an example octokit call.

The search dropdown looks like this on master
image

and like this after the fixes

image

Closes #1475

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -87,8 +87,8 @@ module.exports = {
           OctokitApiMethod: {
             name: node => node.name,
             scope: node => node.scope,
-            route: node => `${node.method} ${node.path}`,
-            method: node => `${node.scope}.${camelCase(node.idName)}`,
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thank you _SO MUCH_ for looking into it _and_ submitting a fix! Much appreciated :)

gr2m:
Works like a charm 👌

PR #1488, opened 2019-09-28
fix(search): remove "octokit-" prefix from markdown searches

### LINKED ISSUE #1446: Documentation search bug
Hi there! 👋 While using the search feature of [the docs](https://octokit.github.io/rest.js/), I noticed that searching for any of the following items results in `octokit-` being added to the url and not leading to the item being searched for.

## Items affected
restjs docs search

## Reproduction
restjs docs search

copperwall:
This fixes a bug where the elasticsearch index always prepends
"octokit-" to the idName of the markdownPage node.

All of the results of the OctokitApiMethod already have octokit-
prefixed in their id field, but the MarkdownRemark nodes don't have the
octokit prefix. By removing the prefix from the MarkdownRemark search
hits, the links in the search results should now correctly point to the
same anchors as the links on the page.

This also doesn't change any existing anchors for the MarkdownPage
links, so this shouldn't break any existing hyperlinks. It should just
correct the MarkdownPage search results.

The difference between the selected links and the other links are that the highlighted links are built from the [allMarkdownRemark query](https://github.com/octokit/rest.js/blob/master/docs/src/pages/index.js#L21) and the others are built from the [allOctokitApiGroup query](https://github.com/octokit/rest.js/blob/master/docs/src/pages/index.js#L35).

Here's the behavior after the fix

Closes #1446

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -76,7 +76,7 @@ module.exports = {
           // For any node of type MarkdownRemark, list how to resolve the fields` values
           MarkdownRemark: {
             title: node => node.frontmatter.title,
-            slug: node => `#octokit-${node.fields.idName}`,
+            slug: node => `#${node.fields.idName}`,
             type: node => "API"
           },
           OctokitApiGroup: {
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -8,7 +8,7 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 
   if (node.internal.type === `MarkdownRemark`) {
     const slug = createFilePath({ node, getNode, basePath: `pages` });
-    const idName = _.kebabCase(node.frontmatter.title);
+    const idName = `${_.kebabCase(node.frontmatter.title)}`;
     createNodeField({
       node,
       name: `slug`,
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -8,7 +8,7 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 
   if (node.internal.type === `MarkdownRemark`) {
     const slug = createFilePath({ node, getNode, basePath: `pages` });
-    const idName = `${_.kebabCase(node.frontmatter.title)}`;
+    const idName = _.kebabCase(node.frontmatter.title);
     createNodeField({
       node,
       name: `slug`,


docs/gatsby-config.js
```diff
@@ -76,7 +76,7 @@ module.exports = {
           // For any node of type MarkdownRemark, list how to resolve the fields` values
           MarkdownRemark: {
             title: node => node.frontmatter.title,
-            slug: node => `#octokit-${node.fields.idName}`,
+            slug: node => `#${node.fields.idName}`,
```
gr2m, line 79: I think in that case we can just do 

```suggestion
            slug: node => node.fields.idName,
```

docs/gatsby-node.js
```diff
@@ -8,7 +8,7 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 
   if (node.internal.type === `MarkdownRemark`) {
     const slug = createFilePath({ node, getNode, basePath: `pages` });
-    const idName = _.kebabCase(node.frontmatter.title);
+    const idName = `${_.kebabCase(node.frontmatter.title)}`;
```
gr2m: why is that change necessary?

gr2m:
Thanks a lot for the pull request! I just left two comments, but overall looks good to me 👍

docs/gatsby-node.js
copperwall: oops sorry about that, my first attempt was to prefix octokit- there but I switched to removing octokit- from the elasticsearch config. I'll remove that.

docs/gatsby-config.js
copperwall: It looks like in the `SearchResults` component it drops in the `slug` value directly into the `href` attribute, so I think we still need the hash prefix in order to link to the anchor. The other slug definitions for `OctokitApiGroup` and `OctokitApiMethod` do this too.

gr2m: I'm sorry I don't follow, I just wonder if we need to wrap `node.fields.idName` into a template string? Or do you refer to something else? I'm lacking a bit context right now, sorry

copperwall: Oh sorry I was talking about how we need the `#` at the start of the template string. `node.fields.idName` doesn't include that so that needs to be prepended to idName to end up with `#previews` or `#usage`.

I'm okay with doing `'#' + node.fields.idName` too.

gr2m: 🤦‍♂ sorry I totally missed the `#`, I better get a coffee 😅

gr2m (approved):
Thank you so much Chris 💐

PR #1491, opened 2019-10-01
refactor: update endpoints

gr2m:
The "Update REST API endpoint methods" workflow now only fetches the endpoints JSON once from https://octokit-routes-graphql-server.now.sh/ and caches the result. The scripts to update code, Typescript and documentation then require the cached JSON instead of fetching their own version.

-----
[View rendered scripts/update-endpoints/generated/README.md](https://github.com/octokit/rest.js/blob/update-endpoints-refactoring/scripts/update-endpoints/generated/README.md)


gr2m (commented):
the typescript CI error is intentional to test the routes update after merging

PR #1492, opened 2019-10-02
fix(search): remove extra octokit prefix in search results

copperwall:
This fixes a bug in the search results caused by #1481 (sorry). The search
results now return the `example` field from `OctokitApiMethod` in its
entirety, which looks something like `octokit.teams.get`. However, the
`SearchResults` component was also prefixing the value with `octokit.`
as well. This lead to results that looked like
`octokit.octokit.teams.get`.

By removing the prefix from SearchResults, this fixes the bug.

Current Behavior:
image

Behavior on this branch:
image

### DIFF
--- a/docs/src/components/search-results.js
+++ b/docs/src/components/search-results.js
@@ -22,7 +22,7 @@ class Row extends PureComponent {
               (<code>{page.route}</code>)
             </small>
             <br />
-            <code>octokit.{page.method}</code>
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Oh wow, I totally missed it, too! Thanks

The Typescript CI error is unrelated, something I work on.

PR #1494, opened 2019-10-04
Show a message for deprecated methods

copperwall:
This shows a general deprecation warning for methods that are deprecated
by GitHub, and a slightly more informational warning for methods that
have been renamed by Octokit.

This adds a new field called `afterId` to the `return` field in the
`OctokitApiMethod` type, which is used for linking to the renamed
method in the documentation.

### Octokit Renamed Method Warning
image

### GitHub Deprecation Warning
image

Since the GitHub deprecation warnings usually already have documentation saying that the endpoint is deprecated, maybe we can highlight that text instead of having a generic error message above that.

I'm still working on this, but I thought I'd open up a PR with the changes so far.


gr2m (approved):
That looks great 👍 Can you rebase on latest master? I think that should resolve the (unrelated) typescript validation error

gr2m:
Thanks so much!

PR #1496, opened 2019-10-05
Add @deprecated jsdoc to types

### LINKED ISSUE #1322: Typescript: add "@deprecated" jsDoc comment to deprecated methods
For example, `.licenses.list()` was deprecated and renamed to `.licenses.listCommonlyUsed()` in [v16.17.0](https://github.com/octokit/rest.js/releases/tag/v16.17.0).

```ts
  licenses: {
    listCommonlyUsed: {
      (params?: Octokit.EmptyParams): Promise<
        Octokit.Response<Octokit.LicensesListCommonlyUsedResponse>
      >;

      endpoint: Octokit.Endpoint;
    };

    list: {
      (params?: Octokit.EmptyParams): Promise<
        Octokit.Response<Octokit.LicensesListResponse>
      >;

      endpoint: Octokit.Endpoint;
    };
```

There should be a deprecation comment above the `list: {` line

copperwall:
This adds some changes to `scripts/update-endpoints/typescript.js` which adds an `@deprecated` jsdoc comment if the method has a truthy `deprecated` field.

This also includes a probably heavy-handed method for marking method params as deprecated by checking if their description includes the word "deprecated" or "Deprecated".

There's a couple of edge cases like

```ts
export type IssuesCreateParams = {
    /**
     * Login for the user that this issue should be assigned to. _NOTE: Only users with push access can set the assignee for new issues. The assignee is silently dropped otherwise. **This field is deprecated.**_
     */
    assignee?: string;
   ...
```

and 

```ts
  export type TeamsUpdateParams = {
    /**
     * The description of the team.
     */
    description?: string;
    /**
     * The name of the team.
     */
    name: string;
    /**
     * The ID of a team to set as the parent team. **Note**: You must pass the `hellcat-preview` media type to use this parameter.
     */
    parent_team_id?: number;
    /**
     * **Deprecated**. The permission that new repositories will be added to the team with when none is specified. Can be one of:
     * \* `pull` - team members can pull, but not push to or administer newly-added repositories.
     * \* `push` - team members can pull and push, but not administer newly-added repositories.
     * \* `admin` - team members can pull, push and administer newly-added repositories.
     */
    permission?: "pull" | "push" | "admin";
   ...
```

I marked this as WIP in case there's a better way to determine if those edge cases above can be marked as deprecated. Would it be better to mark these as renamed or deprecated in `octokit/routes`?

Closes #1322

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21533,13 +21533,13 @@ declare namespace Octokit {
   };
   export type ActivityMarkAsReadParams = {
     /**
-     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: current date/time
+     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: The current timestamp.
      */
     last_read_at?: string;
   };
   export type ActivityMarkNotificationsAsReadForRepoParams = {
     /**
-     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: current date/time
+     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: The current timestamp.
      */
     last_read_at?: string;
 
@@ -22476,11 +22476,38 @@ declare namespace Octokit {
 
     repo: string;
   };
-  export type IssuesCreateParams = {
+  export type IssuesCreateParamsDeprecatedAssignee = {
     /**
      * Login for the user that this issue should be assigned to. _NOTE: Only users with push access can set the assignee for new issues. The assignee is silently dropped otherwise. **This field is deprecated.**_
+     * @deprecated "assignee" parameter has been deprecated and will be removed in future
      */
     assignee?: string;
+    /**
+     * Logins for Users to assign to this issue. _NOTE: Only users with push access can set assignees for new issues. Assignees are silently dropped otherwise._
+     */
+    assignees?: string[];
+    /**
+     * The contents of the issue.
+     */
+    body?: string;
+    /**
+     * Labels to associate with this issue. _NOTE: Only users with push access can set labels for new issues. Labels are silently dropped otherwise._
+     */
+    labels?: string[];
+    /**
+     * The `number` of the milestone to associate this issue with. _NOTE: Only users with push access can set the milestone for new issues. The milestone is silently dropped otherwise._
+     */
+    milestone?: number;
+
+    owner: string;
+
+    repo: string;
+    /**
+     * The title of the issue.
+     */
+    title: string;
+  };
+  export type IssuesCreateParams = {
     /**
      * Logins for Users to assign to this issue. _NOTE: Only users with push access can set assignees for new issues. Assignees are silently dropped otherwise._
      */
@@ -23231,10 +23258,6 @@ declare namespace Octokit {
   };
   export type IssuesUpdateParamsDeprecatedNumber = {
     /**
-     * Login for the user that this issue should be assigned to. **This field is deprecated.**
-     */
-    assignee?: string;
-    /**
      * Logins for Users to assign to this issue. Pass one or more user logins to _replace_ the set of assignees on this Issue. Send an empty array (`[]`) to clear all assignees from the Issue. _NOTE: Only users with push access can set assignees for new issues. Assignees are silently dropped otherwise._
      */
     assignees?: string[];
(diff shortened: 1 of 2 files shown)


copperwall:
Force pushed to add in some prettier changes to make the tests pass ☝️

scripts/update-endpoints/typescript.js
```diff
@@ -120,11 +120,6 @@ async function generateTypes(languageName, templateFile, outputFile) {
             }
 
             const newParamName = entry.params[name].alias;
-
-            if (name === "in_reply_to") {
-              debugger;
-            }
-
```
gr2m, line 127: ooops 😅

index.d.ts
```diff
@@ -21533,13 +21533,13 @@ declare namespace Octokit {
...
     /**
-     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: current date/time
+     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: The current timestamp.
      */
     last_read_at?: string;
   };
   export type ActivityMarkNotificationsAsReadForRepoParams = {
     /**
-     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: current date/time
+     * Describes the last point that notifications were checked. Anything updated since this time will not be updated. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `YYYY-MM-DDTHH:MM:SSZ`. Default: The current timestamp.
```
gr2m, line 21542: Hmm I thought these two changes already existed in `master` 🤔  They are unrelated to your PR. I wonder what happened 🤷‍♂

index.d.ts
```diff
@@ -22476,9 +22476,10 @@ declare namespace Octokit {
 
     repo: string;
   };
-  export type IssuesCreateParams = {
+  export type IssuesCreateParamsDeprecatedAssignee = {
     /**
      * Login for the user that this issue should be assigned to. _NOTE: Only users with push access can set the assignee for new issues. The assignee is silently dropped otherwise. **This field is deprecated.**_
+     * @deprecated "assignee" parameter has been deprecated and will be removed in future
```
gr2m, line 22482: 😍

index.d.ts
```diff
@@ -23303,6 +23327,38 @@ declare namespace Octokit {
...
+    repo: string;
+    /**
+     * State of the issue. Either `open` or `closed`.
+     */
+    state?: "open" | "closed";
+    /**
+     * The title of the issue.
+     */
+    title?: string;
+  };
```
gr2m, line 23361: Where do all these come from suddenly?

index.d.ts
```diff
@@ -25780,10 +25836,6 @@ declare namespace Octokit {
      * The contents of the comment.
      */
     body: string;
-    /**
-     * **Deprecated**. Use **position** parameter instead. Line number in the file to comment on.
-     */
-    line?: number;
```
gr2m, line 25786: Do you know why this was removed? I don't even understand why `IssuesUpdateCommentParams` would have a `line` parameter in the first place, I'm a little confused.

index.d.ts
copperwall: This diff is really weird. When I look at this in an editor, there's a new type for `IssuesUpdateParamsDeprecatedAssignee`, which contains all the fields from `IssuesUpdateParams`, but with an `assignee` field with the deprecated jsdoc.

image

In this diff though it only shows `assignee` and `title` on the `IssuesUpdateParamsDeprecatedAssignee` type.

I can look into why this diff doesn't look like it matches the file content.

index.d.ts
copperwall: This is another weird diff thing. On master the `IssuesUpdateCommentParams` type definition looks like

```ts
  export type IssuesUpdateCommentParams = {
    /**
     * The contents of the comment.
     */
    body: string;

    comment_id: number;

    owner: string;

    repo: string;
  };
```

so I'm not really sure why the diff is showing line as being removed when it doesn't look like was there before.

The line field should be moved to the `ReposCreateCommitCommentParamsDeprecatedLine`, which it shows up in here https://github.com/octokit/rest.js/pull/1496/files#diff-b52768974e6bc0faccb7d4b75b162c99R25865, but that diff is missing some fields like in the `IssuesUpdateParamsDeprecatedAssignee` type. But in the actual file it looks like everything's there.

gr2m:
Overall I think the changes look good, it's just the diff that looks a little weird. If you feel confident that this is all working

I do see the deprecation note when I e.g. try to set assignee for `octokit.issues.create()`

image

And `assignee` is no longer suggested in the typeahead.

image

copperwall:
Yeah, I'm a little confused about that diff too, but I can take some time today to test out the different types that should be deprecated and make sure the odd diff cases like `IssuesUpdateCommentParams` still work correctly.

gr2m:
Sounds great, thanks! Remove the WIP from the PR title once you got through them.

copperwall:
Cool! From what I tested, all the deprecated params don't show up in typeahead and have a `@deprecation` note in their docs.

One thing I noticed is that the `@deprecated` methods still show up in the typeahead when the `@deprecated` params don't. The deprecated methods still have an `@deprecated` note in their documentation, my guess is that it has something to do with how the deprecated params had new deprecated type overloads created and the deprecated methods didn't.

I'm not sure I know enough about TypeScript to know how to hide those fields. Maybe the field can be removed from the namespace object in `index.d.ts`, but I think that would break any TypeScript code that was using one of those deprecated methods.

Or if you're okay with it being in the typeahead as long as the jsdoc has a deprecated attribute then we could merge this as is.

gr2m (approved):
Thanks 👍

PR #1498, opened 2019-10-08
Add timezone option to Octokit constructor

### LINKED ISSUE #1456: Add `timeZone` constructor option
This is a follow up to https://github.com/octokit/rest.js/issues/1455. As `Octokit({headers})` is deprecated, we need to add a new option which sets the time zone header by default for all requests

see also https://developer.github.com/v3/#timezones

copperwall:
This adds a `timeZone` option to the `Octokit` constructor. The value provided is set as the `time-zone` header in the client defaults, so all requests made with that octokit instance will have that `time-zone` header.

I tested this by making a commit on a sandbox repo with the "Europe/Amsterdam" with an octokit instance with a `timeZone` option of "Europe/Amsterdam" and the date information on the commit was made with a timezone of GMT+2, instead of PST.

```git
bash-3.2$ git log -1
commit a5b12ad2038ba1a79a5dd6519db2768eefe0031b (HEAD -> master, origin/master)
Author: Chris Opperwall
Date:   Tue Oct 8 03:39:43 2019 +0200

    super cool commit
bash-3.2$ date
Mon Oct  7 18:45:36 PDT 2019
```

I also added a test case to `smoke-test.js`, since it looked like that's where most of the other options were tested.

I can also start working on the `octokit/core.js` version of this feature. :+1:

Closes #1456

### DIFF
--- a/docs/src/pages/api/00_usage.md
+++ b/docs/src/pages/api/00_usage.md
@@ -38,6 +38,14 @@ Learn more about [API Previews](#previews).
   previews: ['jean-grey', 'symmetra'],
 ```
 
+A default time zone can be enabled by setting the `timeZone` option.
+
+```js
+  timeZone: 'Europe/Amsterdam',
+```
+
+Learn more about [using time zones with the GitHub API](https://developer.github.com/v3/#using-the-time-zone-header).
+
 In order to use Octokit with GitHub Enterprise, set the `baseUrl` option.
 
 ```js
--- a/lib/parse-client-options.js
+++ b/lib/parse-client-options.js
@@ -43,6 +43,10 @@ function parseOptions(options, log, hook) {
     clientDefaults.mediaType.previews = options.previews;
   }
 
+  if (options.timeZone) {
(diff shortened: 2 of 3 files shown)


gr2m (commented):
This looks great, thank you! 

Could you please add it to the usage documentation at https://github.com/octokit/rest.js/blob/master/docs/src/pages/api/00_usage.md? It will show up in the Usage section here: https://octokit.github.io/rest.js/#usage

gr2m (approved):
Perfect, thanks a lot!

gr2m:
The commits are great, too, thanks!

In case you are interested: one thing that I usually do is to push the `test: ..` commit first so it triggers a build, that will give us the red ❌ confirming that the test actually failed. Only then I send the `feat: ...` commit which will get the green ✅ . It helps me to see at the first glance that everything is in order :) No need to do it for a simple fix like this, but it comes in handy at times, especially when we write tests after we wrote the feature / fix

copperwall:
Ohh I see, that makes a lot of sense. Thanks for telling me! I'll definitely make use of that going forward 😄

PR #1504, opened 2019-10-18
Update https-proxy-agent

StephenTG:
npm audit identified a vulnerability in https-proxy-agent that's patched in v3.0.0


gr2m:
let's wait unit v3.0.0 gets promoted to `@latest`, at which point we should get a pullrequest from @greenkeeperio automatically

PR #1507, opened 2019-10-30
fix(typescript): Include done callback in Paginate interface

BinaryMuse:
The `Paginate` interface included in this package has a `callback` property that does not accept the `done` callback that causes pagination to exit early. This PR simply adds those parameters to the interface.


wolfy1339 (approved):

gr2m (approved):
Thanks!

PR #1519, opened 2019-11-26
fix: submitted_at for pr review

maxkomarychev:
original issue: https://github.com/octokit/routes/issues/580


jovel (approved):
@maxkomarychev thanks for making this change 🙏 

re: issue with docs missing this field, it's been prioritized in the docs team queue

gr2m:
Note that this this change will be reverted with the next octokit/routes update unless the according docs got updated. The index.d.ts is fully generated and shouldn’t be edited manually

PR #1538, opened 2019-12-24
improvement: add depreacted fields to tsdoc

pd4d10:


gr2m (approved):
This is great, thank you!

PR #1545, opened 2020-01-13
Type generation

maticzav:
Hi :wave:,

I am creating this PR to track the progress of the type definitions generation from `@octokit/routes`.

I've already implemented the code structure and have rewritten code to `typescript`. What remains is writing a JSON schema to TypeScript processor. I'd be very thankful if you have any suggestions on how I should tackle that.

(cc @gr2m )


gr2m:
Hey @maticzav, thanks for your efforts! 

What problem are you trying to solve here exactly? They TypeScript definitions are already generated for `@octokit/routes` today. And I'm holding off rewriting `@octokit/rest` into TypeScript on purpose, I need to take care of some other things first. I'm working on that.

I'd recommend to create an issue before submitting a pull request in future.

Your PR has a lot of unrelated changes, can you remove these?

maticzav:
Hey @gr2m,

I created this PR as a result of our thread [here](https://github.com/octokit/routes/issues/20#issuecomment-573125761). The problem I am trying to solve is implementing the missing types that the current script doesn't generate.

Currently, the script fetches data from https://octokit-routes-graphql-server.now.sh, but the url isn't working. Type generation relies on examples, even though we have a complete schema in `@octokit/routes`.

Regarding the unrelated changes, could you tell me which ones do you have in mind? I'd be happy to remove them. As a background, I added `.vscode` files because they help you debug the generation process (i.e. we don't have to console log everything to make sense out of it). From what I can tell `prettier` was already installed, but I can remove the configuration. Besides that, there's only changes to `yarn.lock` remaining and ` tsconfig.json`.

I'd be very happy to help you on this, if you'd let me. It's a great library and I use it heavily. 🙂

.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
wolfy1339, line 1: Please remove the edits to this file

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
wolfy1339, line 1: Please remove this file

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
wolfy1339, line 1: Please remove this file

.vscode/tasks.json
```diff
@@ -0,0 +1,10 @@
+{
```
wolfy1339, line 1: Please remove this file

prettier.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
wolfy1339, line 1: Please remove this file, we use the default prettier config

yarn.lock
```diff
@@ -0,0 +1,8658 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
wolfy1339, line 1: Please remove this file

wolfy1339 (changes requested):

gr2m:
> I created this PR as a result of our thread here. The problem I am trying to solve is implementing the missing types that the current script doesn't generate.

Ah sorry I missed the connection

> Currently, the script fetches data from https://octokit-routes-graphql-server.now.sh, but the url isn't working

I'm working on that. For the time being, you can run https://github.com/gr2m/octokit-routes-graphql-server locally and point the script in this repository to load from `http://localhost:4000` until the problem with https://octokit-routes-graphql-server.now.sh is resolved.

gr2m:
https://octokit-routes-graphql-server.now.sh/ is back

maticzav:
https://github.com/octokit/rest.js/commit/dc7f6fec1e026b3e723fc69b5c86b2981668f39f/checks?check_suite_id=398276317 could you check this out? I guess we might not need the PR if the generation script successfully loaded. It seems like it has not done so in quite some time.

gr2m:
I've manually triggered an update, here is the resulting update: https://github.com/octokit/rest.js/pull/1547, I'll review it next

maticzav:
It doesn't seem like it solved the problem. There are still some examples that the scripts don't catch. [`/git/trees`](https://developer.github.com/v3/git/trees/#get-a-tree-recursively), for example, has multiple examples but the script misses. Do you perhaps know what might be causing it?

gr2m:
The [Get a tree recursively](https://developer.github.com/v3/git/trees/#get-a-tree-recursively) section is a special case, I think the docs team is planning to remove it altogether and add the `recursive` query parameter as parameter to [Get a tree](https://developer.github.com/v3/git/trees/#get-a-tree) instead. 

There is currently a workaround implemeted in `octokit/routes`, you can see that the `recursive` parameter is added to the `get-tree.json` operation:

https://github.com/octokit/routes/blob/a5f006c9ed32b95573ca7c4a0e1f15a68e3ce131/openapi/api.github.com/operations/git/get-tree.json#L41-L46

And there is no `get-tree-recursively.json` operation.

What we could maybe do is to try use the response from https://developer.github.com/v3/git/trees/#get-a-tree-recursively in this operation, but it might be tricky. I'd rather talk to the docs team to bump the priority on merging the two if that's a blocker for you.

Anything else that's blocking you right now?

gr2m:
What exactly is the problem with the types for `octokit.git.getTree()`?

maticzav:
I see. Thank you for explaining the background.

The issue I am having is that `getTree` returns `any` instead of a concrete type. Could that be solved by adding an example manually to the generation script?

gr2m:
> The issue I am having is that getTree returns any instead of a concrete type. Could that be solved by adding an example manually to the generation script?

Okay that is indeed odd. I was able to reproduce that behavior. It should be an object with the `sha`, `url`, `tree`, `truncated` properties. So something is not right. 

The response is included in `octokit/routes`: https://github.com/octokit/routes/blob/c60a3186515d37d618b9c728c3aaa45ea6ffa766/openapi/api.github.com/operations/git/get-tree.json#L48-L79

So I assume it's something in the `scripts/update-endpoints/typescript.js` code in this repository. You can run it directly with `node scripts/update-endpoints/typescript.js` and experiment what it could be

maticzav:
@gr2m I've found the issue. The problem is that overrides are not updated. Since `get-tree` merges the recursive and regular endpoint, it requires manual updates. I've added a fix. As you mentioned, however, this is only a temporary fix. If possible, we could separate the two endpoints and resolve the issue altogether.

The PR: https://github.com/octokit/routes/pull/625

gr2m:
does that fix your problem? https://github.com/octokit/rest.js/pull/1556

maticzav:
It does! Thank you for helping me 🙂

gr2m:
released in v16.38.3

PR #1549, opened 2020-01-17
ci: widen branch pattern to cover scoped packages

MaximDevoir:
Packages that are scoped - maybe with an organization name - will not match the current pattern and the workflow will not be triggered.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
       - master
       - next
       - beta
-      - "greenkeeper/*"
+      - "greenkeeper/**"
       - 15.x
       - 16.x # TODO: can we use a pattern for these?
 


gr2m:
Woah good catch, thanks a lot, I wonder if that is something they changed recently? I did it wrong _everywhere_ lol

gr2m (approved):

MaximDevoir:
Haha, I'm not sure.

Probot seems to be in the clear (0 results - [search](https://github.com/search?q=user%3Aprobot+path%3A.github%2Fworkflows+greenkeeper&type=Code)). That's mostly because Probot repositories have yet to migrate to GitHub Actions for testing.

And you've fixed the issue with the other repositories on [Octokit](https://github.com/search?q=user%3Aoctokit+path%3A.github%2Fworkflows+greenkeeper&type=Code), so everything looks alright now. :+1:

PR #1550, opened 2020-01-19
Add support for @octokit/auth

### LINKED ISSUE #1290: Document 2fa sms requirements

gr2m:
part of https://github.com/octokit/rest.js/issues/1546
resolves #1293
resolves #1290

-----
[View rendered docs/src/pages/api/01_authentication.md](https://github.com/octokit/rest.js/blob/octokit-auth/docs/src/pages/api/01_authentication.md)


gr2m:
The `node_modules` folder increased to over 2MB, but it's mostly TypeScript type definitions. I'll update `@octokit/types` to not publish the `scripts/` folder, that will bring it down quite a bit

image

PR #1563, opened 2020-01-29
 App auth strategy

### LINKED ISSUE #13: [Typescript] Octokit "auth" option does not support example from docs for installation tokens
I recently updated the following octokit modules and began receiving a deprecation warning about not providing an `authStrategy`.

## Octokit Modules
```json
"@octokit/app": "4.1.1",
"@octokit/auth-app": "2.4.2",
"@octokit/plugin-retry": "3.0.1",
"@octokit/plugin-throttling": "3.2.0",
"@octokit/rest": "16.40.1",
"@octokit/webhooks": "7.0.0",
```

## Code That Raises Deprecation Warning

```js
import OctokitRest from '@octokit/rest';

const client = new OctokitRest({
    async auth(): Promise<string> {
        const token = await app.getInstallationAccessToken({
            installationId: MY_INSTALLATION_ID,
        });
        return `token ${token}`;
    },
    ...
});
```

## Code I Would Like To Use

```js
import { createAppAuth } from '@octokit/auth-app';
import OctokitRest from '@octokit/rest';

const client = new OctokitRest({
    authStrategy: createAppAuth,
    auth: {
        id: MY_APP_ID,
        privateKey: MY_PRIVATE_KEY,
        installationId: MY_INSTALLATION_ID
    },
    ...
});
```

## Discrepancy to Documentation

Per the [example](https://octokit.github.io/rest.js/#authentication) in the documentation, I should be able to specify the app `id` and `privateKey` as the `auth` settings but those are not valid per the [current typescript definition](https://raw.githubusercontent.com/octokit/rest.js/master/index.d.ts).

image

## Proposal

Not sure if this is on your [roadmap to v17](https://github.com/octokit/rest.js/issues/1546), but if not then here's my two cents.

Since the octokit rest module's [authentication plugin](https://github.com/octokit/rest.js/blob/master/plugins/authentication/index.js#L18) just passes `options.auth` to the auth strategy via `const auth = options.authStrategy(options.auth);`, should the typescript definition for `auth` in this octokit rest module match [StrategyOptions](https://github.com/octokit/auth-app.js/blob/master/src/types.ts#L75) type as shown in the [auth-app usages](https://github.com/octokit/auth-app.js#usage)?

Thanks!

### LINKED ISSUE #1562: [Typescript] Octokit "auth" option does not support example from docs for installation tokens
I recently updated the following octokit modules and began receiving a deprecation warning about not providing an `authStrategy`.

## Octokit Modules
```json
"@octokit/app": "4.1.1",
"@octokit/auth-app": "2.4.2",
"@octokit/plugin-retry": "3.0.1",
"@octokit/plugin-throttling": "3.2.0",
"@octokit/rest": "16.40.1",
"@octokit/webhooks": "7.0.0",
```

## Code That Raises Deprecation Warning

```js
import OctokitRest from '@octokit/rest';

const client = new OctokitRest({
    async auth(): Promise<string> {
        const token = await app.getInstallationAccessToken({
            installationId: MY_INSTALLATION_ID,
        });
        return `token ${token}`;
    },
    ...
});
```

## Code I Would Like To Use

```js
import { createAppAuth } from '@octokit/auth-app';
import OctokitRest from '@octokit/rest';

const client = new OctokitRest({
    authStrategy: createAppAuth,
    auth: {
        id: MY_APP_ID,
        privateKey: MY_PRIVATE_KEY,
        installationId: MY_INSTALLATION_ID
    },
    ...
});
```

## Discrepancy to Documentation

Per the [example](https://octokit.github.io/rest.js/#authentication) in the documentation, I should be able to specify the app `id` and `privateKey` as the `auth` settings but those are not valid per the [current typescript definition](https://raw.githubusercontent.com/octokit/rest.js/master/index.d.ts).

image

## Proposal

Not sure if this is on your [roadmap to v17](https://github.com/octokit/rest.js/issues/1546), but if not then here's my two cents.

Since the octokit rest module's [authentication plugin](https://github.com/octokit/rest.js/blob/master/plugins/authentication/index.js#L18) just passes `options.auth` to the auth strategy via `const auth = options.authStrategy(options.auth);`, should the typescript definition for `auth` in this octokit rest module match [StrategyOptions](https://github.com/octokit/auth-app.js/blob/master/src/types.ts#L75) type as shown in the [auth-app usages](https://github.com/octokit/auth-app.js#usage)?

Thanks!

gr2m:
fixes #1562

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,7 +55,8 @@ declare namespace Octokit {
       | string
       | { username: string; password: string; on2fa: () => Promise<string> }
       | { clientId: string; clientSecret: string }
-      | { (): string | Promise<string> };
+      | { (): string | Promise<string> }
+      | any;
     userAgent?: string;
     previews?: string[];
     baseUrl?: string;
--- a/package-lock.json
+++ b/package-lock.json
@@ -444,9 +444,9 @@
       "integrity": "sha512-ywoxP68aOT3zHCLgWZgwUJatiENeHE7xJzYjfz8WI0goynp96wETBF+d95b8g/uL4QmS6owPVlaxiz3wyMAzcw=="
     },
     "@octokit/plugin-rest-endpoint-methods": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/@octokit/plugin-rest-endpoint-methods/-/plugin-rest-endpoint-methods-2.1.0.tgz",
-      "integrity": "sha512-F2LoDCc4EIfXQlTA29tM2tKOk3QLhbMF5D7zqnD60IZT33PieN1mOSHMd4UPPlrEjgTUZahud6whgdDeM7CrrA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/@octokit/plugin-rest-endpoint-methods/-/plugin-rest-endpoint-methods-2.1.2.tgz",
+      "integrity": "sha512-PS77CqifhDqYONWAxLh+BKGlmuhdEX39JVEVQoWWDvkh5B+2bcg9eaxMEFUEJtfuqdAw33sdGrrlGtqtl+9lqg==",
       "requires": {
         "@octokit/types": "^2.0.1",
         "deprecation": "^2.3.1"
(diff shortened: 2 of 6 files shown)


wolfy1339:
There's one downside to this, at the moment. You will be allowed to pass any value to the argument, all the other types will be ignored

PR #1565, opened 2020-01-30
Add types for the various auth strategies

wolfy1339:
While this isn't perfect, it is the only way to ensure that the types actually correspond to the usage outlined in the docs.

Ref: #1562, #1563

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,6 +18,7 @@ import {
   createTokenAuth,
   createActionAuth
 } from "@octokit/auth";
+import { StrategyInterface } from "@octokit/types";
 
 declare type AuthStrategies =
   | typeof createAppAuth
@@ -64,7 +65,9 @@ declare namespace Octokit {
   export interface EmptyParams {}
 
   export interface Options {
-    authStrategy?: AuthStrategies;
+    authStrategy?:
+      | AuthStrategies
+      | StrategyInterface<any[], any[], any>
     auth?:
       | string
       | { username: string; password: string; on2fa: () => Promise<string> }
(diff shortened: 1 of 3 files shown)


gr2m:
Hmm ... I see two problems with that approach

1. We would need to add the `@octokit/auth-` packages as dependencies, even though only the types are used.
2. It wouldn't allow users to use their own, custom strategies

Any thoughts?

wolfy1339:
1. They really only need to be `devDependencies`. `@octokit/auth` includes them all anyways
2. That seems to be a real blocker here.

Maybe there could a standard way of implementing custom auth strategies that way it makes it easier for the types?

gr2m:
> Maybe there could a standard way of implementing custom auth strategies that way it makes it easier for the types?

Yes there is a standard API overall, the strategy function returns an async `auth` function which also has a `auth.hook` property. The strategy options are different from strategy to strategy though. 

- https://github.com/octokit/auth-token.js accepts a string
- https://github.com/octokit/auth-action.js has no options
- the rest accepts an object

I was hoping that somehow the `auth` option could be derived from the value of `authStrategy`, but doesn't like like TypeScript supports that at this point

wolfy1339:
Everything seems to check out fine for the tests, this should be ready for live testing

gr2m:
I'm working on something else right now, but will check it out next!

gr2m:
@wolfy1339 I'm heads down on v17 of `@octokit/rest`, and I'd rather not merge this into v16 any more. Could you maybe submit the same PR to https://github.com/octokit/core.js? The next `@octokit/rest` version will inherit that automatically

wolfy1339:
I'll take a look to do just that

PR #1576, opened 2020-02-01
fix: artifacts response definition

bokuweb:
fix ` ActionsListWorkflowRunArtifactsResponse`.
I'm trying `octokit.actions.listWorkflowRunArtifacts` however response type mismatch.

### current definition

```
[
  {
    "total_count": 2,
    "artifacts": [...],
  }
]
```

### current real response

```
{
  "total_count": 2,
  "artifacts": [...],
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -23753,10 +23753,6 @@ declare namespace Octokit {
     node_id: string;
     size_in_bytes: number;
   };
-  type ActionsListWorkflowRunArtifactsResponseItem = {
-    artifacts: Array<ActionsListWorkflowRunArtifactsResponseItemArtifactsItem>;
-    total_count: number;
-  };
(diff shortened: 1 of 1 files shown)


gr2m:
The types are generated from the docs, the response is incorrect there: https://developer.github.com/v3/actions/artifacts/#list-workflow-run-artifacts

Once that's fixed, this is going to resolve itself. I'll let the docs team now about it, thank you :)

gr2m:
They are already aware of the problem and are working on it

bokuweb:
@gr2m 
> The types are generated from the docs 

Oops I missed it ... Thanks :)

gr2m:
Absolutely not your fault, I'll make the architecture more clear with the upcoming version :)

You issue should be resolved via https://github.com/octokit/rest.js/pull/1578

Issue #1595, opened 2020-02-09
v17 - documentation


gr2m:
This issue is part of #1546. 

In terms of documentation, everything is up for discussion. Overarching goals are

1. Simple setup
2. Low barrier for documentation contributors
3. Minimal maintenance overhead

## Current state (v16)

- Current website: https://octokit.github.io/rest.js/
- The website is built with [Gatsby](https://www.gatsbyjs.org/), the source code lives at https://github.com/octokit/rest.js/tree/master/docs
- I've created a custom plugin to load nodes for all the REST API endpoint methods, which is probably 95% of the documentation: https://github.com/octokit/rest.js/blob/master/docs/plugins/gatsby-source-octokit-routes/gatsby-node.js
  - The endpoint definitions are coming from https://github.com/octokit/routes. It scrapes [GitHub's REST API docs](https://developer.github.com/v3) and turns it into a JSON API spec
  - I've created a GraphQL API that uses the OpenAP as source, for simpler querying of the endpoints: https://octokit-routes-graphql-server.now.sh/
  - The GraphQL endpoint is used to generate the code for the endpoint methods, the TypeScript definitions, and the docs: https://github.com/octokit/rest.js/tree/master/scripts/update-endpoints
- The website is continuously build and deployed from master

## What is changing (v17)

- `@octokit/rest` v17 will be [roughly 10 lines of code when it's done](https://github.com/octokit/rest.js/blob/beta/src/index.ts). It will be built upon a plugin architecture that I've built out over the past year.

  ```
  @octokit/rest
  ├─┬ @octokit/core@2.4.0
  │ ├── @octokit/auth-token@2.4.0
  │ ├── @octokit/graphql@4.3.1
  │ └─┬ @octokit/request@5.3.1
  │   ├── @octokit/endpoint@5.5.2
  │   └── @octokit/request-error@1.2.1
  ├── @octokit/plugin-paginate-rest@2.0.1
  ├── @octokit/plugin-request-log@1.0.0
  └── @octokit/plugin-rest-endpoint-methods@3.0.0
  ```

- All the REST API endpoint methods are now living in [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js). If possible, I would avoid loading the endpoints definitions again in `@octokit/rest`, to prevent that the `@octokit/rest` documentation gets out of sync with its code.

### Must

- We will continue to maintain v16 for some time. The documentation must remain accessible.

### Should

[...]

PR #1596, opened 2020-02-09
Versioned docs

trevorblades:
This branch implements [`gatsby-source-git`](https://github.com/stevetweeddale/gatsby-source-git) to render pages for older versions of the library. It is meant to help address https://github.com/octokit/rest.js/issues/1595.

- [x] Source content from git remote
- [x] Programatically create pages for older versions
- [x] Add navigation between versions (select?)


docs/gatsby-config.js
```diff
@@ -94,6 +94,15 @@ module.exports = {
...
       }
+    },
+    {
+      resolve: `gatsby-source-git`,
+      options: {
+        name: `v16`,
+        remote: `https://github.com/octokit/rest.js.git`,
+        branch: `master`, // TODO: change to version-16
+        patterns: `docs/src/pages/api/**`
+      }
```
gr2m, line 107: There is a version `15.x`, and I've pushed `16.x` for future v16 maintenance releases.

I'm okay with maintaining this config manually, but if we could make it dynamic in future, that'd be pretty cool.

docs/package.json
```diff
@@ -26,5 +26,8 @@
     "react-typography": "^0.16.18",
     "typography": "^0.16.18",
     "typography-theme-github": "^0.16.19"
+  },
+  "dependencies": {
+    "gatsby-source-git": "^1.0.2"
```
gr2m: I've only used `devDependencies`, as I don't think any gatsby dependency is needed in production?

gr2m:
> Add navigation between versions (select?)

Yeah a select is fine, whatever is simplest to test the functionality

docs/package.json
trevorblades: Ah that's just my habit of installing all my Gatsby things as dependencies. I just moved it over to `devDependencies` 👍

docs/gatsby-config.js
trevorblades: Sounds good! We could implement this similar to how it's done in [Apollo's docs theme](https://github.com/apollographql/gatsby-theme-apollo/blob/master/packages/gatsby-theme-apollo-docs/gatsby-config.js#L99-L111) to make it _more_ dynamic. It wouldn't be automatically detecting what version branches are available, but it allows us to specify versions in a way that is easier to configure.

```js
const versions = [16];

module.exports = {
  plugins: [
    ...versions.map(version => ({
      resolve: 'gatsby-source-git',
      options: {
        name: 'v' + version,
        remote: `https://github.com/octokit/rest.js.git`,
        branch: version + '.x',
        patterns: `docs/src/pages/api/**`
      }
    }))
  ]
};
```

trevorblades:
@gr2m I've just added a select to navigate between versions. You can make changes to the local Markdown files in `src/pages/api` and flip between the current and older versions to verify that it's working!

gr2m:
I'm currently getting an error

image

I've pulled the last commit f01d850, run `npm install`, then `npm run develop`

<details>
<summary>Full logs</summary>

```
r develop

> gatsby-starter-hello-world@ develop /Users/gregor/Projects/octokit/rest.js/docs
> gatsby develop

success open and validate gatsby-configs - 0.061s
success load plugins - 0.558s
success onPreInit - 0.005s
success initialize cache - 0.008s
success copy gatsby files - 0.085s
success onPreBootstrap - 0.014s
success createSchemaCustomization - 0.020s

 ERROR 

localPath is not defined



  ReferenceError: localPath is not defined
  
  - gatsby-node.js:40 getRepo
    [docs]/[gatsby-source-git]/gatsby-node.js:40:59
  
  - task_queues.js:94 processTicksAndRejections
    internal/process/task_queues.js:94:5
  
  - gatsby-node.js:65 async Object.exports.sourceNodes
    [docs]/[gatsby-source-git]/gatsby-node.js:65:12
  

warn The gatsby-source-git plugin has generated no Gatsby nodes. Do you need it?
success source and transform nodes - 0.271s
success building schema - 0.354s

 ERROR #85923  GRAPHQL

There was an error in your GraphQL query:

Cannot query field "allGitRemote" on type "Query".

If you don't expect "allGitRemote" to exist on the type "Query" it is most likely a typo.
However, if you expect "allGitRemote" to exist there are a couple of solutions to common problems:

- If you added a new data source and/or changed something inside gatsby-node.js/gatsby-config.js, please try a restart of your development server
- The field might be accessible in another subfield, please try your query in GraphiQL and use the GraphiQL explorer to see which fields you can query and what shape they have
- You want to optionally use your field "allGitRemote" and right now it is not used anywhere. Therefore Gatsby can't infer the type and add it to the GraphQL schema. A quick fix is to add a least one entry with that field ("dummy content")

It is recommended to explicitly type your GraphQL schema if you want to use optional fields. This way you don't have to add the mentioned "dummy content". Visit our docs to learn how you can define the schema for "Query":
https://www.gatsbyjs.org/docs/schema-customization/#creating-type-definitions

File: gatsby-node.js:63:26


 ERROR #11321  PLUGIN

"gatsby-node.js" threw an error while running the createPages lifecycle:

Cannot read property 'allOctokitApiGroup' of undefined

  111 |     context: {
  112 |       version: ``,
> 113 |       endpointScopes: data.allOctokitApiGroup
      |                            ^
  114 |     }
  115 |   });
  116 | 

File: gatsby-node.js:113:28

failed createPages - 0.065s
success createPagesStatefully - 0.039s
success onPreExtractQueries - 0.002s
success update schema - 0.026s

 ERROR #85923  GRAPHQL

There was an error in your GraphQL query:

Cannot query field "allGitRemote" on type "Query".

If you don't expect "allGitRemote" to exist on the type "Query" it is most likely a typo.
However, if you expect "allGitRemote" to exist there are a couple of solutions to common problems:

- If you added a new data source and/or changed something inside gatsby-node.js/gatsby-config.js, please try a restart of your development server
- The field might be accessible in another subfield, please try your query in GraphiQL and use the GraphiQL explorer to see which fields you can query and what shape they have
- You want to optionally use your field "allGitRemote" and right now it is not used anywhere. Therefore Gatsby can't infer the type and add it to the GraphQL schema. A quick fix is to add a least one entry with that field ("dummy content")

It is recommended to explicitly type your GraphQL schema if you want to use optional fields. This way you don't have to add the mentioned "dummy content". Visit our docs to learn how you can define the schema for "Query":
https://www.gatsbyjs.org/docs/schema-customization/#creating-type-definitions

File: src/components/index-page.js:47:17

failed extract queries from components - 0.171s
warn The GraphQL query in the non-page component "/Users/gregor/Projects/octokit/rest.js/docs/src/components/template.js" will not be run.
Exported queries are only executed for Page components. It's possible you're
trying to create pages in your gatsby-node.js and that's failing for some
reason.

If the failing component(s) is a regular component and not intended to be a page
component, you generally want to use a <StaticQuery> (https://gatsbyjs.org/docs/static-query)
instead of exporting a page query.

If you're more experienced with GraphQL, you can also export GraphQL
fragments from components and compose the fragments in the Page component
query and pass data down into the child component — http://graphql.org/learn/queries/#fragments
success write out requires - 0.015s
success write out redirect data - 0.007s
success onPostBootstrap - 0.002s
⠀
info bootstrap finished - 5.912 s
⠀
success run queries - 0.028s - 1/1 36.35/s
⠀
You can now view gatsby-starter-hello-world in the browser.
⠀
  http://localhost:8000/
⠀
View GraphiQL, an in-browser IDE, to explore your site's data and schema
⠀
  http://localhost:8000/___graphql
⠀
Note that the development build is not optimized.
To create a production build, use gatsby build
⠀
success Building development bundle - 3.803s
success Re-building development bundle - 0.043s
```

</details>

trevorblades:
Hmm... maybe try deleting your `.cache` directory and starting the development server again. That error is happening because `localPath is not defined`. `localPath` is being defined [here](https://github.com/stevetweeddale/gatsby-source-git/blob/72a687281d43e3a6a4d350cb133ac143609644d0/gatsby-node.js#L55-L60) in the source plugin, making me think that it can't find the directory in your cache.

gr2m:
That worked 👍 

Would the next step be to replace the dynamic generation of the endpoint methods docs with static markdown files?

trevorblades:
I've left the dynamic `endpointScopes` [in place for the current version](https://github.com/octokit/rest.js/blob/f01d85062beda262fa1023b9ccbe14717b705daf/docs/gatsby-node.js#L113), but we should definitely generate static markdown files on the `16.x` branch so that the files can be sourced appropriately.

gr2m:
> we should definitely generate static markdown files on the `16.x` branch

I'm on it

gr2m:
I've a PR (https://github.com/octokit/plugin-rest-endpoint-methods.js/pull/33) to add them to the [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js) plugin, e.g. https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/generate-docs-files/docs/users/getAuthenticated.md

For `gatsby-source-git`, do all the files need to be in the current repository's git history? Or can we add a source from a remote repository?

If it's too complicated I'll just generate the same `*.md` files in `@octokit/rest` for now

trevorblades:
We can definitely source files from other repos as long as they're public. You should be able to source content from the `plugin-rest-endpoint-methods.js` repo like this:

```js
{
  resolve: 'gatsby-source-git',
  options: {
    name: 'endpointMethods',
    remote: 'https://github.com/octokit/plugin-rest-endpoint-methods.js',
    pattern: 'docs/**'
  }
}
```

Right now, this branch associates content sourced from a git repo with a particular version. This would introduce some more complexity where we'd source _some_ versioned content from one repo and the rest from another repo and then combine the results. It would also mean that the generated content in `plugin-rest-endpoint-methods.js` would need to follow some version branch scheme as well, right?

My initial reaction would be to just generate the files once, put them in the `16.x` branch, and continue dynamically generating this content for the current version, but I also lack context around how the `endpoint-methods.js` system works. Would those endpoint definitions always apply to all versions? If so, we could just pass along [the dynamic `endpointScopes` variable](https://github.com/octokit/rest.js/blob/f01d85062beda262fa1023b9ccbe14717b705daf/docs/gatsby-node.js#L113) to all pages.

gr2m:
> It would also mean that the generated content in plugin-rest-endpoint-methods.js would need to follow some version branch scheme as well, right?

That should be no problem, they all do already. `@octokit/rest` v16 is depending on `@octokit/rest-endpoint-methods` v2, which also has its corresponding branch `2.x`. I went ahead and merged the PR for the `2.x` branch, I'll do another one for `master`, then `https://github.com/octokit/plugin-rest-endpoint-methods.js.git` should be ready

gr2m:
Ready

- https://github.com/octokit/plugin-rest-endpoint-methods.js/tree/2.x/docs
- https://github.com/octokit/plugin-rest-endpoint-methods.js/tree/master/docs

trevorblades:
Ok awesome! I'll try bringing these files in next 👍

docs/gatsby-config.js
gr2m: Can you add an entry for `beta`? Right now it would just be `16.x` and `beta`. Once version 17 is out, it is going to be `16.x` and `master`.

trevorblades: We only need to add `gatsby-source-git` entries for older versions. The current version is always built from local files. So these changes will be merged into `beta` when they're ready, and then once `beta` is merged to `master`, everything will continue to work without changing anything.

trevorblades:
I [tried sourcing the static markdown files](https://github.com/trevorblades/rest.js/pull/1) that you generated, and it proved to be more challenging than I expected since they are meant to be treated differently than the markdown docs in this repo. This is evident in the way that each type of data (`staticMethods` and `endpointScopes`) are being [rendered in the app](https://github.com/octokit/rest.js/blob/master/docs/src/components/api.js#L98-L109).

Each `endpointScope` is expecting to have an array of methods associated with it, and then we render those methods. If we source them as individual markdown files, they lose the grouping. You may be able to re-group them by exporting some frontmatter with each of the markdown files indicating their name and parent's information, so they can be turned into groups like you're doing in the [existing custom plugin](https://github.com/octokit/rest.js/blob/master/docs/plugins/gatsby-source-octokit-routes/gatsby-node.js).

gr2m:
If it makes things easier, I don't mind changing the structure of the current docs. I'm not a big fan of the huge single page docs anyway. Happy to have one page per `*.md` file, and copy the structure of the markdown file paths in the sources. We can also enforce an `index.md` file or similar for sub folders within the `docs/**` folder.

Anything is on the table. The easier the setup, the better.

trevorblades:
@gr2m I noticed that the generated endpoints markdown is the same for both the `2.x` and `master` branch. Both seem to use the multiple args function signature, rather than the "object of args" style [in the current live docs](https://octokit.github.io/rest.js/#octokit-routes-actions-get-artifact). See:

master: https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/master/docs/actions/getArtifact.md
2.x: https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/2.x/docs/actions/getArtifact.md

Is this intentional?

gr2m:
> Is this intentional?

Ooooops, no! Glad you caught that, I'm fixing that right away

trevorblades:
I managed to get the sourced endpoint data in there! I decided to do that work in [a separate PR](https://github.com/trevorblades/rest.js/pull/1) that I can merge into this one if you agree with the changes.

Notably, I removed the custom Gatbsy plugin that you were using to organize the data previously, and instead grouped the markdown documents by their parent directory [in a GraphQL query](https://github.com/trevorblades/rest.js/blob/e53f7c493c3baab666881daa3143a6a563014667/docs/src/components/template.js#L39-L46).

I deployed the branch on Netlify to make it easier to click around and verify the built version: https://serene-nightingale-a870d8.netlify.com/

gr2m:
Just started reviewing :)

gr2m:
ignore my last comment, it was created automatically by the new `gh` CLI from GitHub 🤷‍♂ I'm just playing around with it. 

Also, I am, in fact, testing it locally now, but still weird for the CLI to leave a comment like that

gr2m:
I wonder if it was possible to always redirect to the `/vXX` slug? That way URLs wouldn't change. Otherwise if someone created a bookmark today, it might no longer work once v18 is out. 

If it's too much trouble I'd make a follow up issue for it though.

trevorblades:
@gr2m I think that would be possible using redirects in Netlify. I'll play around with some settings tonight and get back to you on that!

gr2m:
I should have clarified: I'd want the final page to be statically deployed to GitHub pages, I wouldn't want to rely on Netlify-specific features

trevorblades:
Ok no problem! It would be relatively easy to set the site up to create `/vXX` pages for each version. The part I'm unsure about would be setting up the redirect from `/` to the latest `/vXX` slug. Do you know how that could be done in GH pages?

gr2m:
> Do you know how that could be done in GH pages?

I think for now I'd just do a good old [`<meta http-equiv="refresh" content="0; url=./v16">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#Examples). We might replace it with something more sophisticated in future but that'll do for now.

trevorblades:
@gr2m cool! I've just made a commit that moves the versioned docs logic into a local plugin to make it easier to configure, and queryable in places like the version dropdown, where we'd need to dynamically query the `currentVersion`.

```js
<StaticQuery
  query={graphql`
    {
      # get the current version as defined in gatsby-config.js
      sitePlugin(name: {eq: "gatsby-plugin-versioned-docs"}) {
        pluginOptions {
          currentVersion # very handy 👍
        }
      }
    }
  `}
  render={data => {
    const {currentVersion} = data.sitePlugin.pluginOptions;
    return (
      <select>
        {/* render the current version and map over the others */}
        <option value={currentVersion}>Current ({currentVersion})</option>
        ... other options
      </select>
    );
  }}
/>
```

I've also set up a redirect using the method that you pointed out in the `static/index.html` file. To test these changes, you'll need to check out the branch, `npm run build`, and then `npm run serve`. Otherwise, the redirect in the `index.html` file won't have an effect.

The other significant change here is moving the versioned docs logic into a local plugin, `gatsby-plugin-versioned-docs`. Now, all the versioned docs configuration happens in the root `gatsby-config.js`:

```js
{
  resolve: 'gatsby-plugin-versioned-docs',
  options: {
    currentVersion: 'v17', // configure the path for the current version
    versions: [
      {
        name: 'v16', // the path of the older version
        branch: "16.x", // older versions specify a branch name for this repo
        endpoints: "2.x" // ...and one for the endpoint methods
      }
    ]
  }
}
```

docs/static/index.html
```diff
@@ -0,0 +1 @@
+<meta http-equiv="refresh" content="0;url=/v17">
```
gr2m: Can you think of a way to derive the `/v17` URL from what is set in `gatsby-config.js`'s `options.currentVersion`?

trevorblades: Yup! I just created a page at `src/index.js` that takes the `currentVersion` config option and adds the meta tag within a `Helmet` component. It seems to work just as well, but you still need to `build` and `serve` the site to see the effect locally.

gr2m (approved):
Amazing work, thank you @trevorblades 💐

PR #1625, opened 2020-02-21
fix: Pass type parameters to the Core.plugin method to work around dpoint methods not being defined

wolfy1339:
Without the types, it would pick an arbitrary plugin from the Array and assume the whole array was of that type instead of the more generic and appropriate type which is an Array of OctokitPlugin's

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,19 +2,12 @@ import { Octokit as Core } from "@octokit/core";
 import { requestLog } from "@octokit/plugin-request-log";
 import { paginateRest } from "@octokit/plugin-paginate-rest";
 import { restEndpointMethods } from "@octokit/plugin-rest-endpoint-methods";
+import { OctokitPlugin } from "@octokit/core/dist-types/types"
 
 import { VERSION } from "./version";
 
 export const Octokit = Core
-  // TODO: this really should be
-  //
-  //     .plugin([requestLog, paginateRest, restEndpointMethods])
-  //
-  // but for mystical reasons, using the line above does set the resulting the
-  // `octokit` instance type correctly. Neither `octokit.paginate()` nor all the
-  // endpoint methods such as `octokit.repos.get() are set
-  .plugin(requestLog)
-  .plugin([paginateRest, restEndpointMethods])
+  .plugin<typeof Octokit, OctokitPlugin[]>([requestLog, paginateRest, restEndpointMethods])
   .defaults({
     userAgent: `octokit-rest.js/${VERSION}`
   });


wolfy1339:
Not even this seems to work. But going full on explicit works fine.

gr2m:
Thanks so much for looking into this @wolfy1339! It really bothers me.

Make sure the resulting `octokit` instance has types for all the methods such as `octokit.paginate` and `octokit.repos.get`

I tried to create a minimal test case to reproduce the problem with the plugins array using https://github.com/gr2m/javascript-plugin-architecture-with-typescript-definitions/, but had no luck myself. Maybe you give it a try? Once we have a small reproducible test case, it will be easier to figure it out, and to ask for help

wolfy1339:
From what I have seen in my tests with the codebase, plugins which don't have both arguments seems to "take over" the type that is inferred by typescript

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

gr2m:
> plugins which don't have both arguments

can you clarify what you mean by that? Maybe give an example?

wolfy1339:
I created a more basic test case using only the REST plugins:
```js
import { requestLog } from "@octokit/plugin-request-log";
import { paginateRest } from "@octokit/plugin-paginate-rest";
import { restEndpointMethods } from "@octokit/plugin-rest-endpoint-methods";

const plugins = [
    requestLog,
    paginateRest,
    restEndpointMethods
]
// If you check the type, it will be `(typeof requestLog)[]`
// While the proper type would be (typeof requestLog | typeof paginateRest | typeof restEndpointMethods)[]
```

This seems more like a problem with TypeScript's type interference than anything else.

From what I've seen, it happens in 2 scenarios (or a mix of both):
1. When one includes a plugin that only has the `VERSION` attribute and nothing else.
2. The plugin doesn't take any options at all, and thus only has the one argument: the `Octokit` class.

An example of this is the [`requestLog`](https://github.com/octokit/plugin-request-log.js) plugin, which includes both scenarios.
Another plugin which results in this behaviour is the [`retry`](https://github.com/octokit/plugin-retry.js) plugin, which only has scenario 2

**EDIT**: A plugin that complies with scenario 1 but doesn't exhibit this problem when included in the plugin Array, is the [`paginateRest`](https://github.com/octokit/plugin-paginate-rest.js) plugin.

**EDIT 2**: It seems that it only happens if the return type of the plugin function is `void`, anything else, and it won't make the problem appear

gr2m:
Thanks a lot for looking into it @wolfy1339!

> It seems that it only happens if the return type of the plugin function is void

I'll look into that!

gr2m:
I've tested locally if changing `@octokit/plugin-request-log` to return `{}` instead of `void` would make a difference, but it did not.

> The plugin doesn't take any options at all, and thus only has the one argument

None of the three plugins accept the 2nd `options` argument, so that shouldn't make a difference in that case?

gr2m:
> I've tested locally if changing @octokit/plugin-request-log to return {} instead of void would make a difference, but it did not.

Oh, interesting. When I change `{}` to `{ foo () {} }` then it works!

I've also created a minimal test case to reproduce this problem at https://github.com/gr2m/javascript-plugin-architecture-with-typescript-definitions/pull/15

MaximDevoir:
I'm encountering issues while working on probot/probot#1152 relating to this issue.

It looks like you guys have already discovered that the issue is with the `void` thing.

I'll try to summarize what I see going on within Octokit/rest for myself - maybe it'll help someone else.

The exported `Octokit` file is (see on [unpkg](https://unpkg.com/browse/@octokit/rest@17.0.0/dist-types/index.d.ts)):

```TypeScript
import { Octokit as Core } from "@octokit/core";
export declare const Octokit: (new (...args: any[]) => {
  [x: string]: any;
}) & {
  new (...args: any[]): {
    [x: string]: any;
  };
  plugins: any[];
} & typeof Core  & import("@octokit/core/dist-types/types").Constructor<void> & import("@octokit/core/dist-types/types").Constructor<{
  paginate: import("@octokit/plugin-paginate-rest").PaginateInterface;
} & import("@octokit/plugin-rest-endpoint-methods/dist-types/generated/types").RestEndpointMethods>;
```

The important pieces from above are the plugins being appended to the `Octokit` instance:

```TypeScript
import("@octokit/core/dist-types/types").Constructor<void>
& import("@octokit/core/dist-types/types").Constructor<{
  paginate: import("@octokit/plugin-paginate-rest").PaginateInterface;
}
& import("@octokit/plugin-rest-endpoint-methods/dist-types/generated/types").RestEndpointMethods>
```

This then gets *typed* into:

```TypeScript
Constructor<void>
& Constructor<{
  paginate: PaginateInterface;
}
& RestEndpointMethods
```

which can be simplified as:

```TypeScript
void & { paginate: PaginateInterface } & RestEndpointMethods
```

**End of summarization**

_________

`void` shouldn't be used within an intersection of other objects. Otherwise, users can end up with errors like:

omit-void-type

With Probot, I'm simply omitting the `void` type to avoid issues ATM - albeit in a [very hacky way](https://github.com/MaximDevoir/probot/blob/eca648ebb70cb3e9a8f5bb4256c0a1a19e6489d2/src/github/index.ts#L95-L103).

It's late so I don't have time to test any further now.

I'll check this https://github.com/octokit/core.js/blob/master/src/index.ts#L45-L61 another day. Maybe a type is being inferred incorrectly or is appended to the `Octokit` instance when it shouldn't be.

I'll also check out if the issue is from `@octokit/plugin-request-log`

PR #1633, opened 2020-03-03
docs(usage): Fix example plugin imports

billyvg:
The syntax is wrong and outdated. The plugins used in the example have named exports.

-----
[View rendered docs/src/pages/api/00_usage.md](https://github.com/billyvg/rest.js/blob/patch-1/docs/src/pages/api/00_usage.md)


gr2m (approved):
Great catch, thanks a lot!

Issue #1624, opened 2020-02-21
Upgrading to v17 breaks TypeScript types


willsoto:
Hello! 

Here is the error:

```
'Octokit' refers to a value, but is being used as a type here.
```

```ts
import { Octokit } from "@octokit/rest";

class MyClass {             
  constructor(public logger: Logger, public octokit: Octokit) {}
                                                       ^--- error is here
}
```

I think the typings might just be pointing to the wrong thing somehow? I notice in `package.json` that [`types` refers to `index.d.ts`](https://github.com/octokit/rest.js/blob/master/package.json#L65) but when importing the package it is grabbing the types from `node_modules/@octokit/rest/dist-types/index.d.ts`.

When I open that file, this is all it contains:

```ts
import { Octokit as Core } from "@octokit/core";
export declare const Octokit: (new (...args: any[]) => {
    [x: string]: any;
}) & {
    new (...args: any[]): {
        [x: string]: any;
    };
    plugins: any[];
} & typeof Core & import("@octokit/core/dist-types/types").Constructor<void> & import("@octokit/core/dist-types/types").Constructor<{
    paginate: import("@octokit/plugin-paginate-rest").PaginateInterface;
} & import("@octokit/plugin-rest-endpoint-methods/dist-types/generated/types").RestEndpointMethods>;
```

TypeScript: 3.7.5
Node: 12.15
Yarn: 1.22.0

--- 

This code worked prior to v17 and I don't see anything in the changelog regarding changes to types.

Thanks!

gr2m:
Thank you for reporting the issue Will.

I'm not sure if there is anything I did wrong with the TypeScript setup or if it's a problem with the way the package is built with `@pika/build`.

Before, `@octokit/rest` was written in JavaScript and there was an `index.d.ts` TypeScript file. Starting with version 17, everything is written in TypeScript, so I removed the `index.d.ts` files.

Is there something I could/should change in the TypeScript code to make the exported `Octokit` be both the constructor and the type? It'd probably be needed to change in https://github.com/octokit/core.js.

For the time being you should be able to do `typeof Octokit`. I'll add a note to the changelogs

gr2m:
What's odd is that it seems to work for me, I use the `octokit: Octokit` syntax in all my plugins

https://github.com/octokit/plugin-paginate-rest.js/blob/e7c5443dc4126f9826be9d069c5396b55ef98fc9/src/index.ts#L8-L14

bcoe:
[...]

PR #1638, opened 2020-03-10
fix: load all plugins while preserving type information provided to Octokit

MaximDevoir:
If you're still looking for a *one-liner* solution to load all plugins while preserving the type information provided by other plugins.

Okay, it's not one line after running `lint:fix`, but we do make one `.plugin()` call.

See the related issue and conversation at octokit/core.js#51.

**How does this work?**

Assuming you're familiar with the issue linked above: we're casting the `requestLog` as `() => void`. We still return `void`, but we have removed the parameters that Octokit/core's `.plugin()` method is expecting. This workaround means that TypeScript isn't able to [widen](https://github.com/octokit/core.js/issues/51#issuecomment-596846088) the return values from the plugins in the array - and therefore maintains type information within Octokit using one `.plugin()` call.

octokit-plugins-one-liner

_________

Supersedes and closes #1625 .

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,11 @@ import { restEndpointMethods } from "@octokit/plugin-rest-endpoint-methods";
 import { VERSION } from "./version";
 
 export const Octokit = Core.plugin([
+  // Workaround to prevent TypeScript from widening the inferred return type of
+  // plugins passed to Octokit, which would result in type information (e.g.
+  // methods provided by plugins) not being added to Octokit instances.
+  //
+  // See https://github.com/octokit/core.js/issues/51#issuecomment-596846088
   (requestLog as unknown) as () => void,
   restEndpointMethods,
   paginateRest
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,16 +5,10 @@ import { restEndpointMethods } from "@octokit/plugin-rest-endpoint-methods";
 
 import { VERSION } from "./version";
 
-export const Octokit = Core
-  // TODO: this really should be
-  //
-  //     .plugin([requestLog, paginateRest, restEndpointMethods])
-  //
-  // but for mystical reasons, using the line above does set the resulting the
-  // `octokit` instance type correctly. Neither `octokit.paginate()` nor all the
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -5,16 +5,10 @@ import { restEndpointMethods } from "@octokit/plugin-rest-endpoint-methods";
...
-  // but for mystical reasons, using the line above does set the resulting the
-  // `octokit` instance type correctly. Neither `octokit.paginate()` nor all the
-  // endpoint methods such as `octokit.repos.get() are set
-  .plugin(requestLog)
-  .plugin([paginateRest, restEndpointMethods])
-  .defaults({
-    userAgent: `octokit-rest.js/${VERSION}`
-  });
+export const Octokit = Core.plugin([
+  (requestLog as unknown) as () => void,
```
gr2m, line 14: Oh this is a much more concise workaround, I like it! Could you maybe add a comment to explain what exactly is happening?

We should probably document the workaround for TypeScript users in https://github.com/octokit/rest.js/blob/master/docs/src/pages/api/08_plugins.md, what do you think?

MaximDevoir: > We should probably document the workaround for TypeScript users

I think we should come up with a solution for  octokit/core.js#51 before updating the documentation. Depending on the solution, different documentation changes will be made for both plugin authors and plugin consumers.

gr2m (approved):
Thanks a lot, this is great

JounQin:
@gr2m 

I still have issue with the latest version:

```ts
import { Octokit } from '@octokit/rest'
import { Session as _Session } from 'koa-session'

export interface Session extends _Session {
  user: Octokit.UsersGetAuthenticatedResponse // Cannot find namespace 'Octokit'.
}

export const octokitMap = new Map<number, Octokit>() // 'Octokit' refers to a value, but is being used as a type here.
```

MaximDevoir:
@JounQin This PR doesn't fix that issue (see the related #1624). This PR works around a TS bug that would prevent plugin method type-information from being added to the Octokit instance because of type-widening.

PR #1650, opened 2020-03-19
fix(types): export Octokit type

### LINKED ISSUE #1624: Upgrading to v17 breaks TypeScript types

jamescdavis:
closes #1624

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,3 +17,5 @@ export const Octokit = Core.plugin([
 ]).defaults({
   userAgent: `octokit-rest.js/${VERSION}`
 });
+
+export type Octokit = InstanceType<typeof Octokit>;
--- a/test/integration/smoke.test.ts
+++ b/test/integration/smoke.test.ts
@@ -8,6 +8,11 @@ describe("Smoke tests", () => {
     expect(() => new Octokit()).not.toThrow();
   });
 
+  it("can be used as a type", () => {
+    let octokit: Octokit;
+    octokit = new Octokit();
+  });
+
   it("userAgent option", () => {
     const mock = fetchMock
       .sandbox()


src/index.ts
```diff
@@ -17,3 +17,5 @@ export const Octokit = Core.plugin([
 ]).defaults({
   userAgent: `octokit-rest.js/${VERSION}`
 });
+
+export type Octokit = InstanceType<typeof Octokit>
```
gr2m: could you add a test that fails without this addition, so we don't break that behavior in future?

jamescdavis: It looks like https://github.com/octokit/rest.js/blob/master/test/typescript-validate.ts is no longer checked?

jamescdavis: Oh, I see that was part of the TS rewrite. I'll add it to a jest test.

jamescdavis:
@gr2m ok, I've added a failing test for using `Octokit` as a type (in `smoke.test.ts`, let me know if it should go somewhere else), let that run and fail, and then re-added the fix for green CI.

rwjblue (commented):
Thank you!

gr2m (approved):
Wonderful work, thank you so much!

PR #1663, opened 2020-04-06
Docs: Close nav menu when link is clicked on mobile

copperwall:
This adds a small change to the dropdown nav on the docs site to close the nav when a nav entry is clicked. Previously clicking a nav entry would take you to that anchor/location on the page, but the nav dropdown would still be shown so you would have to click the hamburger button to retract the nav menu.

This also updates the name and description in the `package.json` to not be the default Gatsby starter messages, but I can remove those changes if they're a little too unrelated to the main point of this PR.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "gatsby-starter-hello-world",
-  "description": "Gatsby hello world starter",
+  "name": "octokit-rest.js-docs",
+  "description": "Documentation site for octokit/rest.js",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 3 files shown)


docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "gatsby-starter-hello-world",
-  "description": "Gatsby hello world starter",
+  "name": "octokit-rest.js-docs",
+  "description": "Documentation site for octokit/rest.js",
```
gr2m, line 3: nice catch :)

gr2m (approved):
love the improvement, thanks Chris!

PR #1675, opened 2020-04-18
docs: plugin usage

### LINKED ISSUE #1674: Documentation of plugins mismatches expectations
In most places in the docs it appears the `.plugin` method was expecting to receive the export from a library as a function directly but this syntax doesn't seem to work anymore. For example:

image

But it appears that when requiring a plugin it is wrapped in an object and passing that directly to `.plugin` will work but then causes a failure when you attempt to instantiate the returned object. Perhaps it is from a change in those underlying libraries, but given they are the ones given as examples and probably the most useful probably these docs should be updated.

For instance this code:

```
const { Octokit } = require("@octokit/rest"),
    MyOctokit = Octokit.plugin(require("@octokit/plugin-throttling")),
    octokit = new MyOctokit();
```

Results in:

> TypeError: plugin is not a function

Which is from https://github.com/octokit/core.js/blob/master/src/index.ts#L168 I believe. It looks like one section of the docs was updated, specifically the "Usage" section, see

image

In any case, took me a while to figure out the issue, even if they aren't updated maybe it'll help someone else with the same problem.

gr2m:
thanks @brianjmiller, fixes #1674

-----
[View rendered docs/src/pages/api/09_throttling.md](https://github.com/octokit/rest.js/blob/1674/plugin-usage/docs/src/pages/api/09_throttling.md)
[View rendered docs/src/pages/api/10_retries.md](https://github.com/octokit/rest.js/blob/1674/plugin-usage/docs/src/pages/api/10_retries.md)

### DIFF
--- a/docs/src/pages/api/09_throttling.md
+++ b/docs/src/pages/api/09_throttling.md
@@ -10,7 +10,8 @@ The `throttle.onAbuseLimit` and `throttle.onRateLimit` options are required. Ret
 
 ```js
 const { Octokit } = require("@octokit/rest");
-const MyOctokit = Octokit.plugin(require("@octokit/plugin-throttling"));
+const { throttling } = require("@octokit/plugin-throttling");
+const MyOctokit = Octokit.plugin(throttling);
 
 const octokit = new MyOctokit({
   auth: "token " + process.env.TOKEN,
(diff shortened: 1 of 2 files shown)


flockonus:
@gr2m I'm getting the same error. Is the main doc website still wrong? https://octokit.github.io/rest.js/v18#plugins

gr2m:
That code looks correct, it depends on what is exported by the respective modules. What is your code looking like?

flockonus:
Oh yes, I was having problems with the `throttling` but docs look correct now. 👍

PR #1677, opened 2020-04-18
Docs: Fix whitespace on mobile sizing due to wide content

copperwall:
### Description

This adds a `table-layout: fixed` rule for tables on smaller
screen sizes so that the contents don't overflow the container
element, and adds an `overflow-wrap` rule for long URLs in
code elements.

This also changes the table style in 00_usage.md to use `h3` tags
and `div`s to style the usage snippets so that they don't expand
past the parent container. When using a table with `table-layout: fixed`,
the table header text would be squashed due to the width of the code
snippet part of the table row. If that's too much of a layout change from before I can try out some css grid or flexbox stuff to keep the table look and feel on large screens but then collapse that on smaller screens.

Also there was a small markdown issue where the browser snippet wasn't rendering, this fixes that too.

### Screenshots

Usage Before
image

Usage After
image

Table Before
image

Table After
image

Long URL Before
image

Long URL After
image

### Test Site
https://loving-ride-404dcc.netlify.app/v17

-----
[View rendered docs/src/pages/api/00_usage.md](https://github.com/copperwall/rest.js/blob/fix-whitespace-on-docs/docs/src/pages/api/00_usage.md)

### DIFF
--- a/docs/src/components/layout.css
+++ b/docs/src/components/layout.css
@@ -131,6 +131,17 @@ header > button:focus {
   color: #1ca086;
 }
 
+/* Set table-layout: fixed for small screens where table contents can break out of their container */
+table {
+  table-layout: fixed;
+}
+
+@media (min-width: 55em) {
+  table {
+    table-layout: auto;
+  }
+}
+
 /* Hide the navigation toggle button on big screens, since it’s not needed */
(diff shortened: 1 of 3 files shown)


gr2m (approved):
This is great, thanks so much @copperwall !

PR #1693, opened 2020-04-30
Search Style Follow Ups: Fix version select stretch and remove search duplicates

copperwall:
This fixes an issue where each search entry from the
docs/api/**.md files would show up twice in search results.

This is because we're including versions from v17 and v16, but the
search doesn't specify between them. This adds support to filter based
on the currently selected version.

This also includes style fixes to not stretch the version select
dropdown and a table style fix for endpoint methods to not have the
table extend past its container into the snippet example portion of the
page.

### Screenshots
Non-stretched version select
image

Duplicate search results
Before:
image

After:
image

Table overflow
Before:
image

After:
image

### Demo Site

https://loving-ride-404dcc.netlify.app/v17


gr2m (approved):

gr2m:
Thanks a lot!

PR #1734, opened 2020-05-23
bug: style(docs): fix overflow-y in sections

Alaboudi1:
There is an unnecessary scroll bar on each section header. It is actually confusing as I thought it does something more than just scrolling. This shows up both in chrome and the new edge. Firefox does not have this issue. Here is a picture:
image

This is a simple fix. I tested it on my local machine in Chrome, Firefox and new Edge.


gr2m (approved):
Thanks a bunch 💐

PR #1770, opened 2020-06-17
fix(typescript): return type for `octokit.hook.wrap` callback

gr2m:
fixes bug mentioned in https://github.com/octokit/rest.js/discussions/1769

@douglascayers: could you check if that pull request resolves your problem?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -33704,7 +33704,7 @@ export class Octokit {
           options: Octokit.HookOptions
         ) => Promise<Octokit.Response<any>>,
         options: Octokit.HookOptions
-      ) => void
+      ) => Promise<Octokit.Response<any>>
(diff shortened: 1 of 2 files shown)


gabbloquet (approved):

gr2m:
Merci for the review Gabin!

gabbloquet:
No problem, that's a pleasure

douglascayers:
Thanks @gr2m! -- and sorry I missed your comments that this had been addressed :(

I only now came to check because I was doing some refactoring of our app and came across my comment pointing to https://github.com/octokit/octokit.js/discussions/1769 😅 

For what it's worth... a year later I can confirm this PR did resolve the issue 🎉  

`@octokit/rest@16.43.2`

PR #1893, opened 2020-10-11
Clarify what errors to expect and how the library really works.

webbertakken:
After finally having figured out that there are several authentication scopes and one does not include the other, after ready [this](https://github.com/octokit/octokit.net/blob/main/docs/github-apps.md#authentication) and all related docs several times I managed to implement a method that uses the `installation` scope for a GitHub App.

Some things definitely threw me off at first and I have tried to write the yeast of it as concise as possible.

Hope this will help future implementers.

-----
[View rendered docs/src/pages/api/01_authentication.md](https://github.com/webbertakken/rest.js/blob/patch-1/docs/src/pages/api/01_authentication.md)

### DIFF
--- a/docs/src/pages/api/01_authentication.md
+++ b/docs/src/pages/api/01_authentication.md
@@ -16,6 +16,8 @@ const octokit = new Octokit({
 const { data } = await octokit.request("/user");
 ```
 
+_**Note:** Calling methods outside of the scope of the specific token will result in an `HttpError` with `Bad Credentials` as a message._
+
 To use a different authentication strategy, set `options.authStrategy`. The officially supported authentication strategies are listed on the [`@octokit/auth` README](https://github.com/octokit/auth.js#readme).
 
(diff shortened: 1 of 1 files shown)


oscard0m (approved):

wolfy1339 (approved):

oscard0m:
What's missing to move this PR forward? To update the branch, right?

gr2m (commented):
oh my I made a review but didn't submit 🤦🏼

gr2m:
@wolfy1339 if you don't mind I'd prefer to revert the changes

wolfy1339:
Go ahead

webbertakken:
Would you rephrase it in a better way, as you have a lot more in depth knowledge about it than I. 

The goal was especially to help implementers understand these nuances better.

gr2m:
I took a stab at it here: https://github.com/octokit/rest.js/pull/1897

PR #2080, opened 2021-04-20
fix(README): overall improvements

jeffmcaffer:
The doc is looking good. I found a few typos and wording issues, and have a few suggestions. Summary:

* the pluggability of octokit is exposed in section level bolded text (e.g., **plugin:** @octokit/plugin-rest-endpoint-methods) but the notion of plugins etc is not explained anywhere on the page. Here I suggest turning those references into "links to more doc" or some such and suggest that you add an explicit discussion about plugins somewhere and there, list all the plugins and summarize what they're for. Note also, some places the doc talks about "plugin" and others "standalone method". Not sure if those are the same or different things. In short, for readers of this page, simple is better IMO so remove concepts they don't need and have a separate section/doc that gives the more advanced user more info.
* Some of the changes are a bit pedantic but are trying to be brutally clear whether we're talking about GitHub or Octokit REST or request or ...
* I didn't go through the entire doc. I'm happy to review more but wanted to test the waters with these changes and see if it makes sense.
* There are a few places where I have questions. Will leave comments there once the PR is open

-----
[View rendered README.md](https://github.com/jeffmcaffer/octokit.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ Learn more about [how authentication strategies work](https://github.com/octokit
 
 ### REST API
 
-There are two ways of using the GitHub REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both act on the same GitHub REST API and `octokit.rest` is just a set of convenience methods that use `octokit.request` internally.
+There are two ways of using the GitHub REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both act the same way, the `octokit.rest.*` methods are just added for convenience, they use `octokit.request` internally.
 
 For example
 
@@ -388,9 +388,9 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 
 In both cases a given request is authenticated, retried, and throttled transparently by the `octokit` instance which also manages the `accept` and `user-agent` headers as needed.
 
-`octokit.request` is somewhat lower level and can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
+`octokit.request` can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
 
-#### `octokit.rest` endpoint methods ([`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme))
+#### `octokit.rest` endpoint methods
 
 Every GitHub REST API endpoint has an associated `octokit.rest` endpoint method for better code readability and developer convenience. See [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme) for full details.
 
@@ -407,11 +407,11 @@ await octokit.rest.issues.create({
 
 The `octokit.rest` endpoint methods are generated automatically from [GitHub's OpenAPI specification](https://github.com/github/rest-api-description/). We track operation ID and parameter name changes in order to implement deprecation warnings and reduce the frequency of breaking changes.
 
-Under the covers, every endpoint method is actually an instance of `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
+Under the covers, every endpoint method is just `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
 
-#### Arbitrary requests
+#### `octokit.request()`
 
-You can call the GitHub REST API directly using `octokit.request`. The `reqeust` API matches GitHub's REST API documentation 1:1 so anything you see there, you can call using `request`. See [`@octokit/request`](https://github.com/octokit/request.js#readme) for all the details.
+You can call the GitHub REST API directly using `octokit.request`. The `request` API matches GitHub's REST API documentation 1:1 so anything you see there, you can call using `request`. See [`@octokit/request`](https://github.com/octokit/request.js#readme) for all the details.
 
 Example: [Create an issue](https://docs.github.com/en/rest/reference/issues#create-an-issue)
 
@@ -420,6 +420,7 @@ Example: [Create an issue](https://docs.github.com/en/rest/reference/issues#crea
 The `octokit.request` API call corresponding to that issue creation documentation looks like this:
 
 ```js
+// https://docs.github.com/en/rest/reference/issues#create-an-issue
 await octokit.request("POST /repos/{owner}/{repo}/issues", {
   owner: "octocat",
   repo: "hello-world",
@@ -430,9 +431,9 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 
 The 1st argument is the REST API route as listed in GitHub's API documentation. The 2nd argument is an object with all parameters, independent of whether they are used in the path, query, or body.
 
-#### Pagination ([`@octokit/plugin-paginate-rest`](https://github.com/octokit/plugin-paginate-rest.js/#readme))
+#### Pagination
 
-All `octokit.rest.*.list*` endpoints return the first 30 items by default. If you want to retrieve all items, you an use the pagination API.
+All REST API endpoints that paginate return the first 30 items by default. If you want to retrieve all items, you an use the pagination API. The pagination API expects the REST API route as first argument, but you can also pass any of the `octokit.rest.*.list*` methods for convenience and better code readability.
 
 Example: iterate through all issues in a repository
 
@@ -494,7 +495,7 @@ console.log("topics on octocat/hello-world: %j", data.topics);
 
 Learn more about [Media type formats](https://docs.github.com/en/rest/overview/media-types) and [API previews](https://docs.github.com/en/rest/overview/api-previews).
 
-### GraphQL API queries ([`@octokit/graphql`](https://github.com/octokit/graphql.js#readme))
+### GraphQL API queries
 
 Octokit also supports GitHub's GraphQL API directly -- you can use the same queries shown in the documentation and available in the GraphQL explorer in your calls with `octokit.graphql`.
 
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ Learn more about [how authentication strategies work](https://github.com/octokit
 
 ### REST API
 
-There are two ways of using the REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both work exactly the same, `octokit.rest.*` are just convenience methods that use `octokit.request` internally.
+There are two ways of using the GitHub REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both act on the same GitHub REST API and `octokit.rest` is just a set of convenience methods that use `octokit.request` internally.
 
 For example
 
@@ -386,13 +386,13 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 });
 ```
 
-`octokit.request` can be used to send requests to other domains by passing a full URL. It can also be used to send requests to endpoints that are not (yet) documented on [GitHub's REST API documentation](https://docs.github.com/rest).
+In both cases a given request is authenticated, retried, and throttled transparently by the `octokit` instance which also manages the `accept` and `user-agent` headers as needed.
 
-#### Endpoint methods
+`octokit.request` is somewhat lower level and can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
 
-**plugin**: [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme).
+#### `octokit.rest` endpoint methods ([`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme))
 
-Every REST API endpoint has an according method for better code readability and developer convenience.
+Every GitHub REST API endpoint has an associated `octokit.rest` endpoint method for better code readability and developer convenience. See [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme) for full details.
 
 Example: [Create an issue](https://docs.github.com/en/rest/reference/issues#create-an-issue)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -386,13 +386,13 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 });
 ```
 
-`octokit.request` can be used to send requests to other domains by passing a full URL. It can also be used to send requests to endpoints that are not (yet) documented on [GitHub's REST API documentation](https://docs.github.com/rest).
+In both cases a given request is authenticated, retried, and throttled transparently by the `octokit` instance which also manages the `accept` and `user-agent` headers as needed.
```
jeffmcaffer, line 389: Moved this up because I'm assuming this is true for both rest and request. If not, that's a pretty big differentiator between the two and should be called out

README.md
```diff
@@ -386,13 +386,13 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
...
 
-#### Endpoint methods
+`octokit.request` is somewhat lower level and can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
 
-**plugin**: [`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme).
+#### `octokit.rest` endpoint methods ([`@octokit/plugin-rest-endpoint-methods`](https://github.com/octokit/plugin-rest-endpoint-methods.js/#readme))
```
jeffmcaffer: This will mess with the anchors so you may not want to use this exact form. could just leave it in the text as a link in "See the xxx doc for more details". Or consider making the section heading text be the link text. Depends if you want some formal exposure of the rest/request/graphql component names

README.md
```diff
@@ -405,21 +405,19 @@ await octokit.rest.issues.create({
 });
 ```
 
-The REST API endpoint methods are generated automatically from [GitHub's OpenAPI specification](https://github.com/github/rest-api-description/). We track operation ID or parameter name changes in order to implement deprecation warnings and reduce the frequency of breaking changes.
+The `octokit.rest` endpoint methods are generated automatically from [GitHub's OpenAPI specification](https://github.com/github/rest-api-description/). We track operation ID and parameter name changes in order to implement deprecation warnings and reduce the frequency of breaking changes.
 
-Every endpoint method is an instance of `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
+Under the covers, every endpoint method is actually an instance of `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
 
 #### Arbitrary requests
```
jeffmcaffer: Would be good to find a different name for this. This is really the `request` api but aside from a small number (hopefully zero) of APIs that have not made it into the schema, can I call more things than with `rest`? If so, what?

README.md
```diff
@@ -405,21 +405,19 @@ await octokit.rest.issues.create({
...
 
-**standalone method**: [`@octokit/request`](https://github.com/octokit/request.js#readme)
-
-An http request can be sent using `octokit.request`. It's API is matching GitHub's REST API documentation 1:1.
+You can call the GitHub REST API directly using `octokit.request`. The `reqeust` API matches GitHub's REST API documentation 1:1 so anything you see there, you can call using `request`. See [`@octokit/request`](https://github.com/octokit/request.js#readme) for all the details.
```
jeffmcaffer: If `rest` matches the schema and `request` matches the doc and the doc is driven from the schema, shouldn't they all match?  I get that there may be point in time gaps but in theory the team is looking to drive that window to (near) zero and have the schema be the authority for everything. Given that, how should we position `request`?

Perhaps focusing on "you can cut and past the doc's api URL path directly"?

README.md
```diff
@@ -430,17 +428,9 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 });
 ```
 
-The request is authenticated, retried, and throttled transparently by the `octokit` instance. The `accept` and `user-agent` headers are set as well.
-
 The 1st argument is the REST API route as listed in GitHub's API documentation. The 2nd argument is an object with all parameters, independent of whether they are used in the path, query, or body.
 
-See the standalone [`@octokit/request`](https://github.com/octokit/request.js#readme) module for more details on the `.request()` APIs.
-
-If you only need the request options, but want to send the actual request using your own method, use `octokit.request.endpoint(route, parameters)`. See the standalone [`@octokit/endpoint`](https://github.com/octokit/endpoint.js#readme) module for more details on the `.endpoint()` APIs.
```
jeffmcaffer, line 439: Didn't follow this but it seems "advanced" and fine for folks to discover reading the more detailed doc. If it is important to have here, then we likely need more supporting text to explain the scenario/use case.

README.md
```diff
@@ -430,17 +428,9 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
...
-
-#### Pagination
-
-**plugin**: [`@octokit/plugin-paginate-rest`](https://github.com/octokit/plugin-paginate-rest.js/#readme).
+#### Pagination ([`@octokit/plugin-paginate-rest`](https://github.com/octokit/plugin-paginate-rest.js/#readme))
 
 All `octokit.rest.*.list*` endpoints return the first 30 items by default. If you want to retrieve all items, you an use the pagination API.
```
jeffmcaffer: Does pagination only apply to `rest`? If so, that's a pretty big differentiator that should be called out explicitly here and above to support users deciding which to use.  If not, suggest having two examples, one for `rest` and one for `request` or pointing out that "the `request` form is exactly the same just pass the `request` object to `iterator`)

README.md
```diff
@@ -405,21 +405,19 @@ await octokit.rest.issues.create({
 });
 ```
 
-The REST API endpoint methods are generated automatically from [GitHub's OpenAPI specification](https://github.com/github/rest-api-description/). We track operation ID or parameter name changes in order to implement deprecation warnings and reduce the frequency of breaking changes.
+The `octokit.rest` endpoint methods are generated automatically from [GitHub's OpenAPI specification](https://github.com/github/rest-api-description/). We track operation ID and parameter name changes in order to implement deprecation warnings and reduce the frequency of breaking changes.
 
-Every endpoint method is an instance of `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
+Under the covers, every endpoint method is actually an instance of `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
```
jeffmcaffer: Question": Does every *method* have it's own **instance** of `request`.  Feels like a bit too much implementation detail that leads to that kind of question. What's the main user-facing point you're looking to convey here?

README.md
gr2m: It’s mostly used for 

1. id based routes, which are not yet part of the OpenAPI spec
2. Requests to other domains / other APIs
3. Some people prefer using the request method over the rest endpoint methods because there are fewer changes. Maintenance cost is lower over time

gr2m:
This is great thank you Jeff! I’m busy with other work right now but I’ll get back to your PR with a more thorough review soon

README.md
jeffmcaffer: 1. Got it on the first one. 
2. Not sure I understand the second. What other domains (you mean GHES instances? Shouldn't that be handled automatically somehow?) Other APIs?
3. Completely see the style/aesthetic angle. May need to talk more about changes and maintenance. I would have thought it was the other way around (that `rest` is higher level and so would insulate people from underlying change).

gr2m: > 2. Not sure I understand the second. What other domains (you mean GHES instances? Shouldn't that be handled automatically somehow?) Other APIs?

entirely different APIs. If you build a GitHub app with Octokit but also need to talk to other API endpoints, then the easiest is to use `octokit.request("GET https://api.some-service.com/entities")`

> 3. Completely see the style/aesthetic angle. May need to talk more about changes and maintenance. I would have thought it was the other way around (that `rest` is higher level and so would insulate people from underlying change).

With the rest endpoint methods the `operationId` and the parameters `in: path` are significant. Any change basically results in a breaking change for the rest API. To avoid to many breaking changes, I keep track of them here and declare deprecations: https://github.com/octokit/openapi/tree/main/changes. The deprecations help with migrating to the new method names / parameter names, but just using `ocotkit.request` with a hard coded path prevents you from being bothered by all these changes in the first place

README.md
```diff
@@ -362,7 +362,7 @@ Learn more about [how authentication strategies work](https://github.com/octokit
 
 ### REST API
 
-There are two ways of using the REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both work exactly the same, `octokit.rest.*` are just convenience methods that use `octokit.request` internally.
+There are two ways of using the GitHub REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both act on the same GitHub REST API and `octokit.rest` is just a set of convenience methods that use `octokit.request` internally.
```
gr2m:
```suggestion
There are two ways of using the GitHub REST API, the [`octokit.rest.*` endpoint methods](#endpoint-methods) and [`octokit.request`](#arbitrary-requests). Both act the same way, the `octokit.rest.*` methods are just added for convenience, they use `octokit.request` internally.
```

README.md
```diff
@@ -386,13 +386,13 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
 });
 ```
 
-`octokit.request` can be used to send requests to other domains by passing a full URL. It can also be used to send requests to endpoints that are not (yet) documented on [GitHub's REST API documentation](https://docs.github.com/rest).
+In both cases a given request is authenticated, retried, and throttled transparently by the `octokit` instance which also manages the `accept` and `user-agent` headers as needed.
 
-#### Endpoint methods
+`octokit.request` is somewhat lower level and can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
```
gr2m: I think the `lower level` part is clear, maybe we can leave it out?

```suggestion
`octokit.request` can be used to send requests to other domains by passing a full URL and to send requests to endpoints that are not (yet) documented in [GitHub's REST API documentation](https://docs.github.com/rest).
```

README.md
gr2m: yeah I don't like that change. We can leave out the module name altogether here and mention it in the text
```suggestion
#### `octokit.rest` endpoint methods
```

README.md
gr2m: `request` is not a class, instance is the wrong term here. `request` is a function, but you can create a custom function with defaults by using [`request.defaults()`](https://github.com/octokit/request.js/#requestdefaults).

We can remove the instance wording altogether, what do you think?

```suggestion
Under the covers, every endpoint method is just `octokit.request` with defaults set, so it supports the same parameters as well as the `.endpoint()` API.
```

README.md
gr2m: How about we change the title to just `octokit.request`

```suggestion
#### `octokit.request()`
```

README.md
gr2m: typo 

```suggestion
You can call the GitHub REST API directly using `octokit.request`. The `request` API matches GitHub's REST API documentation 1:1 so anything you see there, you can call using `request`. See [`@octokit/request`](https://github.com/octokit/request.js#readme) for all the details.
```

README.md
```diff
@@ -405,21 +405,19 @@ await octokit.rest.issues.create({
...
+You can call the GitHub REST API directly using `octokit.request`. The `reqeust` API matches GitHub's REST API documentation 1:1 so anything you see there, you can call using `request`. See [`@octokit/request`](https://github.com/octokit/request.js#readme) for all the details.
 
 Example: [Create an issue](https://docs.github.com/en/rest/reference/issues#create-an-issue)
 
 [![Screenshot of REST API reference documentation for Create an issue](assets/create-an-issue-reference.png)](https://docs.github.com/en/rest/reference/issues#create-an-issue)
 
-The according API to send the request with an `octokit` instance:
+The `octokit.request` API call corresponding to that issue creation documentation looks like this:
 
 ```js
```
gr2m, line 422: I like adding the link to the documentation, I think it's a good best practice. What do you think about adding that to the example?

````suggestion
```js
// https://docs.github.com/en/rest/reference/issues#create-an-issue
````

File: README.md [hunk 11]
```diff
@@ -430,17 +428,9 @@ await octokit.request("POST /repos/{owner}/{repo}/issues", {
...
 The 1st argument is the REST API route as listed in GitHub's API documentation. The 2nd argument is an object with all parameters, independent of whether they are used in the path, query, or body.
 
-See the standalone [`@octokit/request`](https://github.com/octokit/request.js#readme) module for more details on the `.request()` APIs.
-
-If you only need the request options, but want to send the actual request using your own method, use `octokit.request.endpoint(route, parameters)`. See the standalone [`@octokit/endpoint`](https://github.com/octokit/endpoint.js#readme) module for more details on the `.endpoint()` APIs.
-
-#### Pagination
-
-**plugin**: [`@octokit/plugin-paginate-rest`](https://github.com/octokit/plugin-paginate-rest.js/#readme).
+#### Pagination ([`@octokit/plugin-paginate-rest`](https://github.com/octokit/plugin-paginate-rest.js/#readme))
```
gr2m:
```suggestion
#### Pagination
```

README.md
gr2m: Yes good idea, let's add two examples.

How about we make it clear in this paragraph that pagination works for both, `the `octokit.rest.*.list*` methods as well as `octokit.request()` for endpoints that paginate, maybe like this?

```suggestion
All REST API endpoints that paginate return the first 30 items by default. If you want to retrieve all items, you an use the pagination API. The pagination API expects the REST API route as first argument, but you can also pass any of the `octokit.rest.*.list*` methods for convenience and better code readability.
```

Then we could add an additional example to the end?

````md
Instead of passing `octokit.rest.issues.listForRepo`, you can pass `"GET /repos/{owner}/{repo}/issues"`, both behave the same way. For example

```js
const issues = await octokit.paginate("GET /repos/{owner}/{repo}/issues", {
  owner: "octocat",
  repo: "hello-world",
  per_page: 100,
});
```
````

File: README.md [hunk 12]
```diff
@@ -504,9 +494,9 @@ console.log("topics on octocat/hello-world: %j", data.topics);
 
 Learn more about [Media type formats](https://docs.github.com/en/rest/overview/media-types) and [API previews](https://docs.github.com/en/rest/overview/api-previews).
 
-### GraphQL API queries
+### GraphQL API queries ([`@octokit/graphql`](https://github.com/octokit/graphql.js#readme))
```
gr2m:
```suggestion
### GraphQL API queries
```

gr2m (commented):
Some feedback and further suggestions, please let me know what you think. Thanks again for helping with the documentation, I really appreciate it, and I really need more help with documentation in general

jeffmcaffer:
That all looks good @gr2m. I updated the PR with your suggestions

gr2m (approved):
Thanks a lot Jeff!

PR #2152, opened 2021-09-15
Add documentation for using  with proxy servers

steve-norman-rft:
Following on from the discussions in https://github.com/octokit/action.js/issues/342, it would make sense to add some notes about using proxy server here.

Hopefully this should cover it.

-----
[View rendered README.md](https://github.com/steve-norman-rft/octokit.js/blob/proxy-docs/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,11 +368,11 @@ By default, the `Octokit` API client does not make use of the standard proxy ser
 For example, this would use a `proxy-agent` generated client that would configure the proxy based on the standard environment variables `http_proxy`, `https_proxy` and `no_proxy`:
 
 ```js
-import ProxyAgent  from "proxy-agent";
+import ProxyAgent from "proxy-agent";
 
 const octokit = new Octokit({
   request: {
-    agent: new ProxyAgent()
+    agent: new ProxyAgent(),
   },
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -360,6 +361,32 @@ await octokit.rest.issues.create({
 
 Learn more about [how authentication strategies work](https://github.com/octokit/authentication-strategies.js/#how-authentication-strategies-work) or how to [create your own](https://github.com/octokit/authentication-strategies.js/#create-your-own-octokit-authentication-strategy-module).
 
+### Proxy Servers
```
gr2m: I think we should make clear that this is only relevant to Node.js, as Octokit is used in other JS runtime environments such as browsers, Deno, and Cloudflare Workers

```suggestion
### Proxy Servers (Node.js only)
```

Other than that it's a great start, we can iterate as more folks run into problems with proxies.

Thanks!

steve-norman-rft: I was thinking about saying something but with a myriad of runtimes I'm not 100% sure what needs it and what doesn't.  Happy to defer that to you though.

gr2m (approved):
Thanks!

PR #2165, opened 2021-10-11
docs: fix deno import

rojvv:
null

-----
[View rendered README.md](https://github.com/rojserbest/octokit.js/blob/patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Deno
 Load <code>octokit</code> directly from <a href="https://cdn.skypack.dev">cdn.skypack.dev</a>
         
 ```ts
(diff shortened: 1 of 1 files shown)


wolfy1339:
While you're at it, would you mind changing the other imports as well?

rojvv:
Do you mean removing `Action` from Node.js imports too?

wolfy1339:
> Do you mean removing `Action` from Node.js imports too?

Yup, that's exactly what I mean

rojvv:
Done, I fixed all of them in a commit.

wolfy1339 (approved):

PR #2181, opened 2021-12-14
docs(README): fix "REST API" section links

TyMick:
Not that the sections are difficult to find from there, but I just noticed that the links are out of date.

-----
[View rendered README.md](https://github.com/TyMick/octokit.js/blob/fix-rest-section-links/README.md)


wolfy1339 (approved):
Thanks alot!

PR #2236, opened 2022-06-26
Fix a typo

marcjansen:
Title says it all.

Please review and thanks for this!

-----
[View rendered README.md](https://github.com/marcjansen/octokit.js/blob/patch-1/README.md)


wolfy1339 (approved):
Nice catch

PR #2243, opened 2022-07-07
ci: stop testing against NodeJS v10, v12

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v10, v12

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,10 @@
         "@pika/plugin-ts-standard-pkg"
       ],
       [
-        "@pika/plugin-build-node"
+        "@pika/plugin-build-node",
+        {
+          "minNodeVersion": "14"
+        }
       ],
       [
         "@pika/plugin-build-web"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,7 +54,10 @@
         "@pika/plugin-ts-standard-pkg"
```
oscard0m, line 54: I'm missing the changes for `.github/workflows/test.yml` updating the matrix of `node_versions`, right?

oscard0m (changes requested):

package.json
oscard0m: Ok, I just saw this: https://github.com/octokit/octokit.js/pull/2244

oscard0m (approved):

wolfy1339:
I merged the other PR into this one, so there is only the one to merge when things are ready

oscard0m (approved):

oscard0m (approved):

PR #2248, opened 2022-07-08
fix(deps): update @octokit dependencies to their latest version

oscard0m:
relates to https://github.com/octokit/octokit.js/pull/2246


wolfy1339:
I bumped other octokit packages to their latest version

We are getting TypeScript errors now:
```
src/app.ts(6,14): error TS2742: The inferred type of 'App' cannot be named without a reference to '@octokit/app/node_modules/@octokit/oauth-app'. This is likely not portable. A type annotation is necessary.
```

wolfy1339 (approved):

PR #2257, opened 2022-07-17
fix(README): `getInstallationOctokit()` doesn't need destructuring

gimenete:
Hey ocktokit team!

I've been playing around with consuming webhooks from a Next.js app so I'm not using probot and I need to create octokit instances manually. I believe this part of the docs is not correct according to the types I'm getting and also taking a look to the source code here https://github.com/octokit/app.js/blob/master/src/get-installation-octokit.ts#L19

Thanks!

-----
[View rendered README.md](https://github.com/gimenete/rest.js/blob/patch-2/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ For example, to implement the above using `App`
 ```js
 const app = new App({ appId, privateKey });
 const { data: slug } = await app.octokit.rest.apps.getAuthenticated();
-const { octokit } = await app.getInstallationOctokit(123);
+const octokit = await app.getInstallationOctokit(123);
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Thanks a lot!

gr2m:
> I'm not using probot and I need to create octokit instances manually

You can also use the `App` constructor:
https://github.com/octokit/octokit.js#app-client

We also export `createNodeMiddleware` which is compatible with Vercel functions:
https://github.com/octokit/octokit.js#webhooks

PR #2266, opened 2022-08-04
fix: use "secondary" instead of "abuse" for rate limit

nimadini:
GitHub renamed `abuse limits` to `secondary rate limits` in 2021:
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits

`@octokit/plugin-throttling` followed suit in March 2022, and released
version `3.6.0` with this new name (and deprecated `onAbuseLimit`):
octokit/plugin-throttling.js#455

This commit updates the codebase to use the new name.

-----
[View rendered README.md](https://github.com/nimadini/octokit.js/blob/main/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ The time zone header will determine the timezone used for generating the timesta
 
 `Octokit` implements request throttling using [`@octokit/plugin-throttling`](https://github.com/octokit/plugin-throttling.js/#readme)
 
-By default, requests are retried once and warnings are logged in case hitting a rate or abuse limit
+By default, requests are retried once and warnings are logged in case of hitting a rate or secondary rate limit
 
 ```js
 {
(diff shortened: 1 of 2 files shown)


wolfy1339:
Could you follow the same changes as in `@octokit/plugin-throttling`, and add an alias for the old name that logs a deprecation message.

gr2m:
> Could you follow the same changes as in `@octokit/plugin-throttling`, and add an alias for the old name that logs a deprecation message.

I think the deprecation message should already be shown, as the options are simply passed through to `@octokit/plugin-throttling`. Can you confirm @nimadini?

nimadini:
Thanks for the quick review!

I confirmed the deprecation message (from `@octokit/plugin-throttling`) is printed when a secondary rate limit is hit.

To test it, I pointed my `Octokit` client to a dummy server that returns a `403` with an error message that has the `secondary rate` keyword in it. (See https://gist.github.com/nimadini/516b82ff69e374259ce23909c1a6f6da for details.)

Maybe an earlier place  `@octokit/plugin-throttling` could warn users is during construction of the `Octokit` client, instead of later on when the secondary rate limit is hit. Specially because by design `@octokit/plugin-throttling` implements [GitHub's best practices](https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-rate-limits) to avoid hitting the secondary rate limit in the first place (and it does a great job at that). So in practice, this warning message may never be shown to the user. 🙂

nimadini:
Confirmed above @gr2m. 🙂 Can you please take another look when you have a minute?

gr2m (approved):

PR #2322, opened 2022-10-04
docs(MAINTAINING): initial version

gr2m:
@nickfloyd this includes notes on maintenance releases

-----
[View rendered CONTRIBUTING.md](https://github.com/octokit/octokit.js/blob/create-maintaining-docs/CONTRIBUTING.md)
[View rendered MAINTAINING.md](https://github.com/octokit/octokit.js/blob/create-maintaining-docs/MAINTAINING.md)

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -1,0 +1,91 @@
+# Maintenance releases
+
+## Merging the Pull Request & releasing a new version
+
+Releases are automated using [semantic-release](https://github.com/semantic-release/semantic-release).
+The following commit message conventions determine which version is released:
+
+1. `fix: ...` or `fix(scope name): ...` prefix in subject: bumps fix version, e.g. `1.2.3` → `1.2.4`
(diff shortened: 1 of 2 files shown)


nickfloyd (approved):
Thank you for doing this! ❤️

MAINTAINING.md
```diff
@@ -0,0 +1,91 @@
...
+    branches:
+      - main
+      - next
+      - beta
+      - "*.x"
+```
+
+### 1. Create a branch for the maintenance version
+
+Find the latest version that was released on the maintenance version. For example, if the current version is 3.1, and you want to release maintenance versions for 2.x, then find the latest 2.x version. Say that's 2.10.9. In that example, create a tag based on this tag
```
wolfy1339, line 54: Change `In that example, create a tag based on this tag` to `In that example, create a branch based on this tag`
There's a typo there, it should be `create a branch` and not `create a tag`

PR #2379, opened 2023-01-12
Enable CodeQL

kfcampbell:
Part of ensuring that each repo of ours has a build and CodeQL CI check. 

----

### Before the change?

* There was no CodeQL scanning. 

### After the change?

* CodeQL scanning is implemented on the repository. 

### Other information

* N/A

----

### Pull request checklist
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Added the appropriate label for the given change

- [ ] Yes (Please add the `Type: Breaking change` label)
- [X] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

Please add the corresponding label for change this PR introduces:
- Bugfix: `Type: Bug`
- Feature/model/API additions: `Type: Feature`
- Updates to docs or samples: `Type: Documentation`
- Dependencies/code cleanup: `Type: Maintenance`

----

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", beta ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '19 5 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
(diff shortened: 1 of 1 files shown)


nickfloyd (commented):
@kfcampbell I think you will need to add an exclusion for app.test.ts given it uses a mocked bearer token.

kfcampbell:
I've played with it a little bit and here's where I've landed:

- First, I experimented with substituting obviously fake values for the bearer token, private key, etc. but nixed the idea as I believe it would require significant test restructuring
- Then I looked into seeing if there's a declarative way to ignore this specific line, which [there isn't](https://github.com/github/codeql/issues/9298)
- I started looking into ignoring the whole file, but decided instead to use the UI to dismiss this alert. That way this won't happen again but we'll be notified if there's other issues in the file. 

Thoughts?

nickfloyd (approved):

PR #2446, opened 2023-05-13
ci: stop testing against NodeJS v14, v16

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v14, v16

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,9 +13,8 @@ jobs:
     strategy:
       matrix:
         node_version:
-          - 14
-          - 16
           - 18
+          - 20
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node_version }}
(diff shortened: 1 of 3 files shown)


kfcampbell:
@wolfy1339 Node 16 is still supported until [September 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol). What's the justification for stripping it before then?

wolfy1339:
It was suggested by @gr2m in https://github.com/octokit/webhooks.js/pull/849#pullrequestreview-1425468491

In order to avoid making another breaking release in a couple months

kfcampbell (approved):

wolfy1339:
Something is up with branch protection. It currently requires node 10 and 12. which were removed a whiles back

gr2m:
> Something is up with branch protection. It currently requires node 10 and 12. which were removed a whiles back

most likely we didn't update the `beta` branch protections, on it

PR #2456, opened 2023-05-22
feat: v3

wolfy1339:
BREAKING CHANGE: Drop support for NodeJS v14, v16
BREAKING CHANGE: Remove previews support for the REST API
BREAKING CHANGE: remove `agent` option from `octokit.request()`
BREAKING CHANGE: Replace support for Node.js http(s) Agents with documentation on using fetch dispatchers instead (via `@octokit/request`)
BREAKING CHANGE: Remove ability to pass custom request options, except for `method`, `headers`, `body`, `signal` (via @octokit/request)


styfle (commented):
Also need to bump `@octokit/core` to `5.0.0`, right?

wolfy1339:
This is the very last PR to get merged, as this depends on practically every other Octokit module.

nickfloyd (approved):

PR #2457, opened 2023-05-23
Add additional thoughts on rolling out breaking changes

kfcampbell:
Today we talked about how to roll out breaking changes safely in the octokit.js ecosystem. This PR captures that discussion in a more durable place.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -23,7 +23,7 @@ If the pull request looks good but does not follow the commit conventions, use t
 
 ## Breaking changes
 
-When merging a breaking change using the PR body as specified above, extra care must be taken. Breaking changes must first be merged into the `beta` branch, where further testing may be conducted and additional breaking changes may be combined before cutting a release. After changes are combined and tested, merge the `beta` branch into `main` as documented above. `beta` branches are intended to be short-lived.
+When merging a breaking change using the PR body as specified above, extra care must be taken. Breaking changes must first be merged into the `beta` branch, where further testing may be conducted and additional breaking changes may be combined before cutting a release. The `beta` branch should be created for this purpose and based on `main`. Once the first commit to the branch is landed, a draft PR should be created from `beta` to `main` with the title vX, where X is the next major version. After changes are combined and tested, mark the PR as ready for review, get it reviewed, and merge the `beta` branch into `main` as documented above. `beta` branches are intended to be short-lived.
 
 Note the repository for the change: if it's dependent on other repos where the same change must be made, merge the leaf nodes first before the nodes higher up the tree. Your merge order should look something like:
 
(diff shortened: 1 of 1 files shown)


wolfy1339:
It may be good to note that for changes requiring type changes, `@octokit/types` needs to be updated

MAINTAINING.md
```diff
@@ -21,6 +21,22 @@ If the pull request looks good but does not follow the commit conventions, use t
...
+1. plugins
+1. auth strategies
+1. core
+1. \*-methods
+1. oauth-app
+1. webhooks
+1. app
+1. octokit/octokit.js
+1. octokit/rest.js
+
```
kfcampbell, line 42: Do we (or should we) have a procedure for how often we diff the `beta` branch with `main` and create new releases?

gr2m: Usually the two shouldn't diverge, `beta` branches should be short lived. Or is your question about something else?

gr2m (approved):

MAINTAINING.md
wolfy1339: I don't think there is a procedure, however one should create a draft PR for the next major version merging `beta` into `main`, ex: https://github.com/octokit/request.js/pull/586 and then mark it ready for review when it is ready

kfcampbell: In that case, should I amend the procedure to contain something like: 

"After the breaking changes are combined and tested in the `beta` branch, merge the `beta` branch into `main` using a PR with "BREAKING CHANGE" in the body as documented above. `beta` branches are intended to be short-lived."

kfcampbell:
Good idea @wolfy1339! Would that go at the top of the chain since it's a dependency of many other modules?

wolfy1339:
That sounds right

MAINTAINING.md
kfcampbell: I'ved added similar language to the PR; please let me know what you think!

kfcampbell:
I've updated the list accordingly!

MAINTAINING.md
gr2m: Good pony by @wolfy1339, we should create beta pull requests as drafts until all planned breaking changes have been merged into it

gr2m: Maybe add that

1. beta branches should be deleted first and re-created based on main
2. Once the first pull request is merged into beta, create a draft pull request from beta to main with the title vX where X is the next breaking version.
3 Once all breaking changes are merged into beta, test and set the pull request to ready for review

nickfloyd (approved):

MAINTAINING.md
kfcampbell: Sounds good! I've now done so.

PR #2674, opened 2024-05-07
feat: v4

wolfy1339:
Rebased #2665
Closes https://github.com/octokit/octokit.js/pull/2665

BREAKING CHANGE: package is now ESM
BREAKING CHANGE: remove authentication type "oauth" that was previously deprecated
BREAKING CHANGE: Switch from `@octokit/webhooks-types` to `@octokit/openapi-webhooks-types` due to the transition to using GitHub's OpenAPI schema

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,19 +9,19 @@
       "version": "0.0.0-development",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


gr2m (approved):
Let's goooo

PR #2835, opened 2025-04-08
Update @octokit/plugin-throttling

jonathanmorley:
To pick up this fix https://github.com/octokit/plugin-throttling.js/pull/779

Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

- [ ] Yes
- [ ] No

----


wolfy1339:
Thanks, but this isn't needed. Since the update is in range, users can simply update their local npm packages and it should be picked up automatically.

Also, next time you send a PR for dependency updates, make sure to update the `package-lock.json`

